PR #1, opened 2021-04-23
service discovery module proposal idea

gytis:
Just a proposal for the service-discovery module organization. If you agree with the idea, I'll complete the implementation and also do the same with the load balancer

### DIFF
--- a/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceDiscovery.java
+++ b/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceDiscovery.java
@@ -1,0 +1,10 @@
+package io.smallrye.discovery;
+
+import java.util.List;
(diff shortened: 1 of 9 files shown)


service-discovery/api/src/main/java/io/smallrye/discovery/ServiceDiscovery.java
```diff
@@ -0,0 +1,10 @@
+package io.smallrye.discovery;
+
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+public interface ServiceDiscovery {
```
michalszynkiewicz, line 6: I thought of this entity as something that would provide `ServiceInstance`s for a specific service name

michalszynkiewicz:
I let myself merge it, we can refactor further later

PR #2, opened 2021-04-28
Reorganise

gytis:
This is a second part of the reorganization POC containing a client side load balancer API its round-robin implementation.

### DIFF
--- a/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceInstance.java
+++ b/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceInstance.java
@@ -1,14 +1,20 @@
 package io.smallrye.discovery;
 
 public class ServiceInstance {
-    // TODO we probably need a service name too
-    final long id;
+    private final String serviceName;
 
-    final String value;
+    private final long id;
 
-    public ServiceInstance(long id, String value) {
+    private final String value;
+
+    public ServiceInstance(String serviceName, long id, String value) {
+        this.serviceName = serviceName;
         this.id = id;
         this.value = value;
+    }
+
+    public String getServiceName() {
+        return serviceName;
     }
 
     public long getId() {
--- a/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceInstance.java
+++ b/service-discovery/api/src/main/java/io/smallrye/discovery/ServiceInstance.java
@@ -1,20 +1,13 @@
 package io.smallrye.discovery;
 
 public class ServiceInstance {
-    private final String serviceName;
-
     private final long id;
 
     private final String value;
 
-    public ServiceInstance(String serviceName, long id, String value) {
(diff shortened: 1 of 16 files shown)


service-discovery/api/src/main/java/io/smallrye/discovery/ServiceInstance.java
```diff
@@ -1,16 +1,22 @@
...
-    final long id;
+    private final String serviceName;
 
-    final String value;
+    private final long id;
 
-    public ServiceInstance(long id, String value) {
+    private final String value;
+
+    public ServiceInstance(String serviceName, long id, String value) {
```
michalszynkiewicz, line 10: My idea was for `serviceName` not to be here, so that a user would define service discovery separately for different services. It seems to me that with `serviceName` here every usage of service discovery would have to be configured with both, the strategy name and serviceName

michalszynkiewicz: E.g.
```
smallrye.service-discovery.mydiscovery.type=kubernetes
smallrye.service-discovery.mydiscovery.service-name=my-kube-service-name


# and then when configuring load balancer:
# this could maybe be skipped if name matches service-discovery name
smallrye.load-balancer.my-balancer.service-discovery=mydiscovery 
smallrye.load-balancer.my-balancer.type=round-robin
```

then the integration with e.g. resteasy would be a bit simpler because this code would only call
```
loadBalancer.getServiceInstance()
```

gytis: But the user still would need to know which load balancer to get. For me this seems basically the same as knowing which service instance to get, because in this case load-balancer is matched to a service 1-to-1.

So you either do `loadBalancer.getInstance("service-name")` or `provider.getLoadBalancer("service-name").getInstance()`. Except in the first case CDI injection would be easier because there would be only one load-balancer

gytis: As for this specific class, it is a service instance, so I think it makes sense for it to know to which service group it belongs.

michalszynkiewicz: I think it's either `provider.getLoadBalancer(<some-name>)` vs `provider.getLoadBalancer(<some-name>).getInstance(<service-name>)`, isn't it?

gytis: My thinking was that a load balancer could be either received by:
```
LoadBalancer loadBalancer = provider.loadBalancer();
```
or 
```
@Inject
LoadBalancer loadBalancer;
```

I.e. there would be one load balancer for all the services. Rather than a single load balancer per service where user would need to inject a provider every time.

PR #5, opened 2021-06-02
Reimplement service discovery

gytis:
I've tweaked the codebase according to the discussions in the last week's call. If you'd like I'd be happy to have a quick call to discuss it.


gytis:
cc @cescoffier

PR #11, opened 2021-06-18
simplify stuff

michalszynkiewicz:
added abstraction for config, simplified creating new service discoveries and load balancers


api/src/main/java/io/smallrye/dux/Dux.java
```diff
@@ -0,0 +1,104 @@
...
+        return providers.stream().map(ServiceLoader.Provider::get)
+                .collect(Collectors.toMap(ElementWithType::type, Function.identity()));
+    }
+
+    private static final Dux dux = new Dux();
+
+    public static Dux getInstance() {
+        return dux;
+    }
+}
```
aureamunoz, line 104: I think we should add some unit tests in the api module, wdyt? May be in a differente PR?

api/src/main/java/io/smallrye/dux/spi/ServiceDiscoveryProvider.java
```diff
@@ -0,0 +1,10 @@
+package io.smallrye.dux.spi;
```
aureamunoz, line 1: What means `spi`?

aureamunoz (approved):
A minor remark and a question about unit tests in the `api` module but, in general, looks good to me

api/src/main/java/io/smallrye/dux/spi/ServiceDiscoveryProvider.java
michalszynkiewicz: spi is for people who implement new service discovery and load balancer strategies. It stands for Service Provider Interface

PR #14, opened 2021-06-21
Config provider for MicroProfile Config

michalszynkiewicz:


microprofile/src/test/java/io/smallrye/dux/microprofile/TestMpConfigProvider.java
```diff
@@ -0,0 +1,38 @@
+package io.smallrye.dux.microprofile;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigBuilder;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+
+public class TestMpConfigProvider extends ConfigProviderResolver {
```
aureamunoz, line 7: Can we rename this to `TestMicroProfileProvider`?

aureamunoz (changes requested):
Just a minor remark

microprofile/src/test/java/io/smallrye/dux/microprofile/TestMpConfigProvider.java
michalszynkiewicz: I renamed to TestMicroProfileConfigProvider

PR #16, opened 2021-06-22
Some documentation

michalszynkiewicz:


aureamunoz (approved):
LGTM

README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Configuring user applications
+The project's core aims to be vendor neutral. Aside from using Mutiny for the APIs,
+it has no dependencies. As such, it provides a universal abstraction of configuration 
+and vendors can choose their own way of providing Dux configuration.
+
+MicroProfile based frameworks can use MicroProfile Config to configure Dux. 
+Below is an example snippet of such a configuration:
+```properties
+dux.my-service.service-discovery=consul
```
Ladicek, line 21: Would it be better to use a reserved config key (such as `type` or `kind`) here? Reason is that translation of this to the YAML format, which seems to be quite popular, is not exactly intuitive:

```yaml
dux:
  my-service:
    service-discovery:
      ~: consul
      some-service-discovery-property: property-value
```

MP Reactive Messaging for example uses reserved config keys, and it seems to work.

README.md
```diff
@@ -0,0 +1,79 @@
...
+MicroProfile Config support for configuring SmallRye Dux is provided by the 
+`smallrye-dux-microprofile` artifact.
+
+## Integrating client libraries with Dux
+Client libraries would most often use a `LoadBalancer` to get a single address
+for a call.
+`LoadBalancer` for a service can be retrieved from `Dux` singleton:
+
+```java
+LoadBalancer loadBalancer = Dux.getInstance().getLoadBalancer(serviceName);
```
Ladicek, line 39: Did you think of CDI integration? Being able to `@Inject @ServiceName("my-service") LoadBalancer` (or `@Inject @ServiceName("my-service") ServiceDiscovery` would be nice.

michalszynkiewicz: That should be easy to do :)

The classes that use it would usually be quite low-level, I didn't think it's likely they would be CDI beans.

README.md
michalszynkiewicz: good point, it seemed unnecessary to add a key here for properties, I didn't think of YAML.

WDYT of changing this to `dux.my-service.service-discovery.type=consul` @aureamunoz ?

aureamunoz: Yes, I agree with adding the key. However I don't like very much `type`, what about `tool`? 
`dux.my-service.service-discovery.tool=consul`

michalszynkiewicz: it's not always a tool, it can be e.g. `static` which means a list of values provided in config by users

PR #46, opened 2021-08-04
gathering statistics for calls

michalszynkiewicz:
This change adds an API to gather statistics of calls and a (not well tested, probably not optimal) statistics based load balancer.
The purpose of the load balancer is to validate the API of gathering the information, not to provide a production-ready solution


aureamunoz (approved):
LGTM

PR #69, opened 2021-09-14
renamed stat-based load balancer to least-response-time

michalszynkiewicz:


Ladicek:
As a nitpicker, I have to point out that the package name is `responsetime`, and the class name is `LeastLatency`. Latency and response time, while often used interchangeably, are actually two different things :-)

michalszynkiewicz:
> As a nitpicker, I have to point out that the package name is `responsetime`, and the class name is `LeastLatency`. Latency and response time, while often used interchangeably, are actually two different things :-)

good catch, will fix it

michalszynkiewicz:
@Ladicek fixed

Ladicek:
Ah sorry, I wasn't clear. What you measure is actually response time, not latency :-) But then, as I said, the two terms are often used interchangeably, so as long as the codebase is internally consistent, it's fine. Nitpicker logging off :laughing:

michalszynkiewicz:
> Ah sorry, I wasn't clear. What you measure is actually response time, not latency :-) But then, as I said, the two terms are often used interchangeably, so as long as the codebase is internally consistent, it's fine. Nitpicker logging off

Hmmm... at first I just wanted to make it consistent. But maybe I should rename to `least-response-time`. It's not that long and closest to true (only omits the fact that we treat errors even worse than long response times)

michalszynkiewicz:
@cescoffier how do you like `least-response-time`?

cescoffier:
+1

aureamunoz (approved):

PR #74, opened 2021-09-16
don't initialize Stork in a static block

michalszynkiewicz:


michalszynkiewicz:
@aureamunoz could you do a post-merge review?

PR #95, opened 2021-10-11
Rewrite the readme and provide a contribution guide

cescoffier:


michalszynkiewicz:
thanks @cescoffier ! can you leave the part about extending/customizing Stork until we have a proper page with this info?

PR #120, opened 2021-10-25
Initial proposal for service instance metadata

cescoffier:
Many service discovery mechanism allows service instances to be decorated with metadata. For example, Kubernetes has labels; Consul has tags, and so on. 

This metadata can be helpful in the service selection phase (load balancer) to implement fine-grain selection/ranking of the instances. 

At the moment, Stork does not provide metadata. This PR is a proposal to expose metadata on the ServiceInstance.

The proposal is the following:

- Metadata are `Map<String, Object>`
- Metadata are immutable; it should not be possible to update the Map during the selection, for example
- If no metadata is provided, an empty Map is used.

The PR extends the Consul and Kubernetes service discovery with metadata support. The Consul service discovery copies the consul id, node, and tags to the Map, using custom key sets. I'm not a big fan of this, as it seems to introduce some brittle named-based API. For Kubernetes, I use the endpoint labels (if any).


api/src/main/java/io/smallrye/stork/DefaultServiceInstance.java
```diff
@@ -8,21 +10,36 @@
...
         return host;
     }
 
+    @Override
     public int getPort() {
         return port;
     }
+
+    @Override
+    public ServiceMetadata getMetadata() {
```
cescoffier, line 42: I would name it ServiceInstanceMetadata for clarity. As there is a `Service` class.

api/src/main/java/io/smallrye/stork/ServiceMetadata.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Map;
+
+public class ServiceMetadata {
+
+    private final Map<String, String> labels;
+    //TODO rename
+    private final Map<String, Object> metadata;
+
+    public ServiceMetadata(Map<String, String> labels, Map<String, Object> metadata) {
```
cescoffier, line 11: What's the rationale behind these two lists? Aren't labels a list of strings that can be stored in metadata, or keys with a specific "singleton" value?

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscovery.java
```diff
@@ -11,17 +9,24 @@
...
 import io.vertx.core.Vertx;
 import io.vertx.ext.consul.ConsulClient;
 import io.vertx.ext.consul.ConsulClientOptions;
+import io.vertx.ext.consul.Service;
 import io.vertx.ext.consul.ServiceEntry;
 import io.vertx.ext.consul.ServiceEntryList;
 
 public class ConsulServiceDiscovery extends CachingServiceDiscovery {
 
+    public static final String META_CONSUL_SERVICE_ID = "consul-service-id";
```
cescoffier, line 25: The main question is how do we communicate these keys? They are kind of an API that cannot be discovered.

api/src/main/java/io/smallrye/stork/ServiceMetadata.java
aureamunoz: In K8s labels are key-value pairs we were leaning towards making a special case for them, the `Map<String, String> labels`. The other stuff is not common to all serviceDiscovery implementations so it seemed better to me to wrap them in the Map<String,Object> at the moment and see how it goes

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscovery.java
aureamunoz: I've just pushed a version with labels and metadata in the ServiceInstance, I will work on this now

aureamunoz:
Just pushed the refactor to have the labels and metadata as separated Maps. I need now work in making consul keys discovery able and tests but I think it should be cool to have this PR merged as it is starting to be rebased onto master to many times.... @michalszynkiewicz @cescoffier

PR #135, opened 2021-11-04
A new approach to handle the service instance refresh.

cescoffier:
This PR proposes a new approach to handle the service instance "refresh". It reuses the idea of the timed cache we had before but uses the wonderful `memoize` operator that Julien did. It removes a lot of the code that was required before and simplifies how the unis are created.

CC @michalszynkiewicz @jponge


api/src/main/java/io/smallrye/stork/CachingServiceDiscovery.java
```diff
@@ -45,73 +41,26 @@ public CachingServiceDiscovery(ServiceDiscoveryConfig config) {
                     period,
                     e);
         }
+
+        this.lastResults = Collections.emptyList();
+        this.instances = Uni.createFrom().deferred(() -> fetchNewServiceInstances(this.lastResults)
+                .invoke(l -> this.lastResults = l)
+                .onFailure().invoke(this::handleFetchError)
+                .onFailure().recoverWithItem(this.lastResults))
+                .memoize().atLeast(this.refreshPeriod);
```
michalszynkiewicz, line 50: does atLeast literally mean at least? What does it mean in practice?

michalszynkiewicz:
nice!

api/src/main/java/io/smallrye/stork/CachingServiceDiscovery.java
jponge: It memoizes for at least the given amount of time, then invalidates.

michalszynkiewicz: `atLeast` is misleading for me then. But the method is a great fit for what we need :)

jponge: Would `duringAtLeast` have worked better?

jponge: I mean, `Uni::memoize()` is still marked as an experimental API so... 🤣

michalszynkiewicz: I read `at least` for `time ∈ [this, +∞)`.
What about `forPeriod`?

jponge: @michalszynkiewicz let's move to a Mutiny discussion: https://github.com/smallrye/smallrye-mutiny/discussions/categories/ideas

michalszynkiewicz: done

PR #138, opened 2021-11-08
Add metadata and labels to ServiceInstance for helping the load balancer

aureamunoz:
As disused together I added two structures to the `ServiceInstance` for storing labels and other stuff which could be useful for selecting the instance to use. Finally I added a extra layer `Metadata` wrapping the `Map<String, Object>` because it adds a meaningfull value I think.
This is a Draft because the `Metadata` object can still be improved and overall it needs to be documented also but want to have already your thoughts.

### DIFF
--- a/api/src/main/java/io/smallrye/stork/Metadata.java
+++ b/api/src/main/java/io/smallrye/stork/Metadata.java
@@ -1,41 +1,108 @@
 package io.smallrye.stork;
 
+import java.util.Collections;
 import java.util.EnumMap;
 import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
 
+/**
+ * Service Instance metadata.
+ * <p>
+ * This class stores service instance metadata that can be used by the load balancer to select the service instance to use.
+ * <p>
+ * Instances of this class are <strong>immutable</strong>. Modification operations return new instances.
+ * <p>
+ * You can creates new instances using the {@link #of(Class)}, {@link #of(Class, Map)} and {@link #with(Enum, Object)}methods.
+ * <p>
+ */
 public class Metadata<T extends Enum<T>> {
 
-    private EnumMap<T, Object> metatada;
-    Class<T> enumType;
+    private final EnumMap<T, Object> metatada;
+    private final Class<T> clazz;
+    private static final Metadata EMPTY = new Metadata(DefaultMetadataKey.class, Collections.emptyMap());
 
-    public Metadata(Class<T> key) {
-        this.metatada = new EnumMap<T, Object>(key);
-        this.enumType = key;
+    private Metadata(Class<T> key, Map<T, Object> metatada) {
+        if (metatada.isEmpty()) {
+            this.metatada = new EnumMap<T, Object>(key);
+            this.clazz = key;
+        } else {
+            this.metatada = new EnumMap<T, Object>(metatada);
+            this.clazz = key;
+        }
     }
 
-    public Metadata() {
-
+    /**
+     * Returns an empty set of metadata.
+     *
+     * @return the empty instance
+     */
+    public static Metadata empty() {
+        return EMPTY;
     }
 
-    public Map<T, Object> getMetadata() {
+    /**
+     * Returns an instance of {@link Metadata} containing metadata values.
+     *
+     * @param metadata the metadata, must not be {@code null}, must not contain {@code null},
+     *        must not contain multiple objects of the same class
+     * @return the new metadata
+     */
+    public static Metadata of(Class<?> key, Map<?, Object> metadata) {
+        if (metadata == null) {
+            throw new IllegalArgumentException("`metadata` must not be `null`");
+        }
+        return new Metadata(key, metadata);
+    }
+
+    /**
+     * Returns an instance of {@link Metadata} containing an empty set of values.
+     *
+     * @param key the type of metadata, must not be {@code null}
+     * @return the new metadata
+     */
+    public static Metadata of(Class<?> key) {
+        if (key == null) {
+            throw new IllegalArgumentException("`key` must not be `null`");
+        }
+        return new Metadata(key, Collections.emptyMap());
+    }
+
+    /**
+     * Creates a new instance of {@link Metadata} with the current entries, plus {@code item}.
+     * If the current set of metadata contains already an instance of the class of {@code item}, the value is replaced
+     * in the returned {@link Metadata}.
+     *
+     * @param item the metadata to be added, must not be {@code null}.
+     * @return the new instance of {@link Metadata}
+     */
+    public Metadata with(T key, Object item) {
+        if (key == null) {
+            throw new IllegalArgumentException("`key` must not be `null`");
+        }
+        if (item == null) {
+            throw new IllegalArgumentException(key.name() + " should not be `null`");
+        }
+        EnumMap<T, Object> copy = new EnumMap<>(metatada);
+        copy.put(key, item);
+        return new Metadata(this.clazz, copy);
+    }
+
+    public EnumMap<T, Object> getMetadata() {
         return metatada;
     }
 
-    public Optional<Object> get(T key) {
-        if (key == null) {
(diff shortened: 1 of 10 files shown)


service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscovery.java
```diff
@@ -1,7 +1,11 @@
 package io.smallrye.stork.servicediscovery.consul;
 
+import static io.smallrye.stork.servicediscovery.consul.ConsulMetadataKey.*;
```
cescoffier, line 3: someone told me that this is bad as the _death star_ :-D

cescoffier (changes requested):
It goes in the right direction. I still fail to see how the value type is communicated. Shouldn't MetadataKey also convey the "type" of the value? Also, it does not improve discoverability. Maybe this is a lost cause, and only documentation would allow that.

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscovery.java
aureamunoz: Gr gr gr, it slipped through the cracks

aureamunoz:
> It goes in the right direction. I still fail to see how the value type is communicated. Shouldn't MetadataKey also convey the "type" of the value? Also, it does not improve discoverability. Maybe this is a lost cause, and only documentation would allow that.

I gave a new try in the last commit, It's the same idea but wrapping the EnumMap in the metadata, this simplifies the code when putting values in it. However in the testing side is not great .....wdyt?

aureamunoz:
cc @michalszynkiewicz

cescoffier:
That looks a LOT better! 
Now we can discover the metadata.

api/src/main/java/io/smallrye/stork/Metadata.java
```diff
@@ -0,0 +1,41 @@
...
+    private EnumMap<T, Object> metatada;
+    Class<T> enumType;
+
+    public Metadata(Class<T> key) {
+        this.metatada = new EnumMap<T, Object>(key);
+        this.enumType = key;
+    }
+
+    public Metadata() {
+
```
cescoffier, line 19: Why do we have this constructor? Is it to create empty metadata? In this case, I would use a specific static method and hide this constructor. 

The problem is that accesses to the metadata will throw NPE.

api/src/main/java/io/smallrye/stork/Metadata.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public class Metadata<T extends Enum<T>> {
+
+    private EnumMap<T, Object> metatada;
+    Class<T> enumType;
```
cescoffier, line 11: should be final.

api/src/main/java/io/smallrye/stork/Metadata.java
```diff
@@ -0,0 +1,41 @@
+package io.smallrye.stork;
+
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public class Metadata<T extends Enum<T>> {
+
+    private EnumMap<T, Object> metatada;
```
cescoffier, line 10: should be final.

api/src/main/java/io/smallrye/stork/Metadata.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public Optional<Object> get(T key) {
+        if (key == null) {
+            throw new IllegalArgumentException("key must not be `null`");
+        }
+        return Optional.ofNullable(metatada.get(key));
+    }
+
+    public void put(T key, Object o) {
+        metatada.put(key, o);
```
cescoffier, line 34: I'm wondering if we should not add all entries in the constructor and be totally immutable.

api/src/main/java/io/smallrye/stork/Metadata.java
```diff
@@ -0,0 +1,41 @@
+package io.smallrye.stork;
+
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
```
cescoffier, line 7: Javadoc missing

api/src/main/java/io/smallrye/stork/ServiceInstance.java
```diff
@@ -31,8 +31,8 @@
     /**
      * @return the metadata of the instance, empty if none.
      */
-    default Map<String, Object> getMetadata() {
-        return Collections.emptyMap();
+    default Metadata<? extends MetadataKey> getMetadata() {
+        return new Metadata();
```
cescoffier, line 35: Access to that metadata will fail, because the map is `null`.

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulMetadataKey.java
```diff
@@ -0,0 +1,22 @@
+package io.smallrye.stork.servicediscovery.consul;
+
+import io.smallrye.stork.MetadataKey;
+
+public enum ConsulMetadataKey implements MetadataKey {
+
+    META_CONSUL_SERVICE_ID("consul-service-id"),
```
cescoffier, line 7: What I like about this approach is that we can discover the metadata, and even document them :-D

service-discovery/consul/src/test/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryIT.java
```diff
@@ -14,17 +17,15 @@
...
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
 import org.testcontainers.utility.DockerImageName;
 
-import io.smallrye.stork.Service;
-import io.smallrye.stork.ServiceInstance;
-import io.smallrye.stork.Stork;
-import io.smallrye.stork.StorkTestUtils;
+import io.smallrye.stork.*;
```
cescoffier, line 28: Star import.

service-discovery/consul/src/test/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryIT.java
```diff
@@ -121,7 +131,8 @@ void shouldRefetchWhenRefreshPeriodReached() throws InterruptedException {
         Thread.sleep(5000);
```
cescoffier, line 131: Unrelated to that PR - can we use awaitility to wait until the set change?

cescoffier (changes requested):

api/src/main/java/io/smallrye/stork/Metadata.java
aureamunoz: Yes, I needed something for using when I don't know the actual type, a kind of default....ok, I will do it with a static method

michalszynkiewicz:
In this design, there is no way to share a metadata key between service discovery providers, right?
Is this intentional, do we want to pull out each parameters that can be common the way we did with labels?

aureamunoz:
I refactored this according to the comments of the review, I think I have taken everything into account except the @michalszynkiewicz comment about sharing some metadata between service providers, what did we say about it? 
Anyway I wanted to push the last version so you can take a look.
For documentation, I plan to add docs about this and caching service discovery may be in a new PR, wdyt? 
cc @cescoffier

michalszynkiewicz (approved):

cescoffier (approved):

PR #162, opened 2021-12-02
Infrastructure class

### LINKED ISSUE #117: How to handle Vert.x instances - add Infrastructure class
We now have multiple service discovery providers using a Vert.x instance. At the moment each provider creates its own instance. There are a few problems with this approach:

* the instances are not closed - of course, Stork is intended to be closed when the JVM terminates, but in multi-app environments, it could be a problem;
* each instance has a cost, and lead to the creation of event loop threads

We would need a way for the providers to get access to a potentially managed instance of Vert.x to avoid the two issues.

michalszynkiewicz:
fixes #117

### DIFF
--- a/api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
+++ b/api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
@@ -1,0 +1,22 @@
+package io.smallrye.stork.integration;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+/**
+ * An implementation of {@link StorkInfrastructure} that creates utility objects from {@code defaultSupplier}
+ * and caches them
+ */
+public class DefaultStorkInfrastructure implements StorkInfrastructure {
+    private final Map<Class<?>, Object> utilities = new ConcurrentHashMap<>();
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T get(Class<T> utilityClass, Supplier<T> defaultSupplier) {
+        Objects.requireNonNull(utilityClass, "utilityClass cannot be null");
+        Objects.requireNonNull(defaultSupplier, "defaultSupplier cannot be null");
+        return (T) utilities.computeIfAbsent(utilityClass, key -> defaultSupplier.get());
+    }
+}
--- a/api/src/main/java/io/smallrye/stork/integration/StorkInfrastructure.java
+++ b/api/src/main/java/io/smallrye/stork/integration/StorkInfrastructure.java
@@ -1,0 +1,27 @@
+package io.smallrye.stork.integration;
+
+import java.util.function.Supplier;
+
+/**
+ * A provider for "utility" objects used by service discovery and load balancer implementations.
+ * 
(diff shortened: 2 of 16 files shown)


michalszynkiewicz:
how quarkus integration looks like with this: https://github.com/michalszynkiewicz/quarkus/commit/820006edae65a0987b434cf5f4ee8d02afed348f

cescoffier:
I like the approach.
The danger is that beans you will look up in Quarkus may have been removed. So, extensions will be required to be sure that the beans are not removed.

api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
```diff
@@ -0,0 +1,15 @@
+package io.smallrye.stork.integration;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+public class DefaultStorkInfrastructure implements StorkInfrastructure {
```
cescoffier, line 7: Javadoc missing

api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
```diff
@@ -0,0 +1,15 @@
...
+    private final Map<Class<?>, Object> utilities = new ConcurrentHashMap<>();
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T get(Class<T> utilityClass, Supplier<T> defaultSupplier) {
+        return (T) utilities.computeIfAbsent(utilityClass, key -> defaultSupplier.get());
```
cescoffier, line 13: Probability check that utilityClass and defaultSupplier are not null.

api/src/main/java/io/smallrye/stork/integration/StorkInfrastructure.java
```diff
@@ -0,0 +1,7 @@
+package io.smallrye.stork.integration;
+
+import java.util.function.Supplier;
+
+public interface StorkInfrastructure {
+    <T> T get(Class<T> utilityClass, Supplier<T> defaultSupplier);
```
cescoffier, line 6: Javadoc missing

It should specify what should happen if the default supplier throws a runtime exception

cescoffier (changes requested):

aureamunoz (approved):
LGTM

api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
michalszynkiewicz: added

api/src/main/java/io/smallrye/stork/integration/DefaultStorkInfrastructure.java
michalszynkiewicz: done

api/src/main/java/io/smallrye/stork/integration/StorkInfrastructure.java
michalszynkiewicz: I added the javadoc. What happens with exceptions thrown by supplier depends on an implementation...

cescoffier (approved):

PR #163, opened 2021-12-02
avoid NPE during tests

aureamunoz:
When only the service discovery strategy is configured without any other setting, a NPE happens. Fix for aligning with the prod behaviour, the only thing mandatory is the service discovery and load balancer, for the rest, Stork provides defaults values.


microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java
```diff
@@ -78,18 +78,18 @@ public MicroProfileConfigProvider() {
             String loadBalancerType = properties.get(LOAD_BALANCER);
             builder.setServiceName(serviceName);
             if (loadBalancerType != null) {
-                SimpleServiceConfig.SimpleLoadBalancerConfig loadBalancer = new SimpleServiceConfig.SimpleLoadBalancerConfig(
```
michalszynkiewicz, line 81: I like the old names better TBH

michalszynkiewicz (approved):

PR #164, opened 2021-12-02
Add description about the server authentication in the k8s SD

aureamunoz:
…ce discovery strategy

### DIFF
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -17,6 +17,10 @@ First, you need to add the Stork Kubernetes Service Discovery provider:
 </dependency>
 ```
 
+####A few words about server authentication. 
+Stork uses Fabric8 Kubernetes Client to access the Kubernetes resources, concretely the `DefaultKubernetesClient` implementation. It will try to read the ~/.kube/config file in your home directory and load information required for authenticating with the Kubernetes API server. If you are using DefaultKubernetesClient from inside a Pod, it will load ~/.kube/config from the ServiceAccount volume mounted inside the Pod. You can override this configuration if you want a more complex configuration.
+
+
 ## Configuration
 
 For each service expected to be exposed as _Kubernetes Service_, configure the lookup:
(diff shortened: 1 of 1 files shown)


cescoffier (approved):

docs/kubernetes.md
```diff
@@ -31,10 +35,9 @@ Stork looks for the _Kubernetes Service_ with the given name (`my-service` in th
...
 
-
 Supported attributes are the following:
 
 | Attribute              | Mandatory    | Default Value     | Description                                                         |
 |-- -------------------- | ------------ | ----------------- | ------------------------------------------------------------------- |
 | `k8s-host`             | No           | _master url_      | The Kubernetes API host                                             |
-| `k8s-namespace`        | No           |                   | The namespace of the service. Use `all` to discover all namespaces. |
\ No newline at end of file
+| `k8s-namespace`        | Yes          |                   | The namespace of the service. Use `all` to discover all namespaces. |
```
michalszynkiewicz, line 43: right now if not specified, it defaults to the namespace that is in the .kube/config or the mounted volume, right?

aureamunoz: Yes

michalszynkiewicz: so adding it in the properties is not required

aureamunoz: I should have elaborate more. After [this modification](https://github.com/smallrye/smallrye-stork/blob/30df8133581fee56b8b5282dfb89a2424c686379/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java#L52), we need to have it in the properties.

cescoffier: still, that does not make it mandatory.

michalszynkiewicz: @aureamunoz it's not needed to have it in properties when it is configured in the environment, then we get it from https://github.com/smallrye/smallrye-stork/blob/30df8133581fee56b8b5282dfb89a2424c686379/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java#L49

PR #165, opened 2021-12-02
add an option to specify service discovery and LB type in properties

### LINKED ISSUE #22: reconsider adding `type` to the load balancer and service discovery properties
Right now the type of service discovery for a service is configured as follows:
```properties
dux.my-service.service-discovery=consul
dux.my-service.service-discovery.some-service-discovery-property=property-value
```

As @Ladicek pointed out, it is not good for YAML where users will end up with:
```yaml
dux:
  my-service:
    service-discovery:
      ~: consul
      some-service-discovery-property: property-value
```

I think we need to go back to:
```properties
dux.my-service.service-discovery.type=consul
dux.my-service.service-discovery.some-service-discovery-property=property-value
```

michalszynkiewicz:
fixes #22


aureamunoz (approved):
LGTM

PR #182, opened 2021-12-24
fix: add pod labels to service instances

### LINKED ISSUE #181: Service instances doesn't contain the labels from the endpoints backend pods
In Kubernetes service discovery strategy we should retrieve and add the labels from the backend pods for a given service, they could be needed during a custom load balancing strategy. Currently, only the service labels are being added to the service instances.

aureamunoz:
Fixes #181

### DIFF
--- a/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
+++ b/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
@@ -5,8 +5,10 @@ import static io.smallrye.stork.servicediscovery.kubernetes.KubernetesMetadataKe
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -15,6 +17,7 @@ import io.fabric8.kubernetes.api.model.EndpointAddress;
 import io.fabric8.kubernetes.api.model.EndpointPort;
 import io.fabric8.kubernetes.api.model.EndpointSubset;
 import io.fabric8.kubernetes.api.model.Endpoints;
+import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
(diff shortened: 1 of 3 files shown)


cescoffier (approved):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -66,25 +69,44 @@ public KubernetesServiceDiscovery(String serviceName, ServiceDiscoveryConfig con
 
     @Override
     public Uni<List<ServiceInstance>> fetchNewServiceInstances(List<ServiceInstance> previousInstances) {
-        Uni<List<Endpoints>> endpointsUni = Uni.createFrom().emitter(
+        Uni<Map<Endpoints, List<Pod>>> endpointsUni = Uni.createFrom().emitter(
                 emitter -> {
                     vertx.executeBlocking(future -> {
-                        List<Endpoints> endpoints = new ArrayList<>();
+                        Map<Endpoints, List<Pod>> items = new HashMap<>();
+                        List<Pod> backendPods = new ArrayList<>();
```
michalszynkiewicz, line 76: it seems to me we're collecting pods for an  endpoint in this collection, shouldn't it be initialized in the for loop in line 81?

michalszynkiewicz (changes requested):
I'm afraid the pod collection is shared between services. Isn't it?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Yes, good catch. I refactored to have it in the last commit

michalszynkiewicz (approved):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -116,18 +143,24 @@ public KubernetesServiceDiscovery(String serviceName, ServiceDiscoveryConfig con
...
                     } else {
-                        Map<String, String> labels = endPoints.getMetadata().getLabels() != null
+                        Map<String, String> labels = new HashMap<>(endPoints.getMetadata().getLabels() != null
                                 ? endPoints.getMetadata().getLabels()
-                                : Collections.emptyMap();
+                                : Collections.emptyMap());
+                        Map<String, String> podLabels = pods.stream().filter(pod -> pod.getMetadata().getName().equals(podName))
+                                .findFirst().get().getMetadata().getLabels();
+                        for (Map.Entry<String, String> label : podLabels.entrySet()) {
+                            labels.putIfAbsent(label.getKey(), label.getValue());
```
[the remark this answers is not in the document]
aureamunoz: Only one pod can have a given name at a time.

PR #185, opened 2022-01-03
Use round-robin as default load-balancing strategy

cescoffier:


cescoffier:
Converting to draft - the service loader mock causes too many issues... Need to find a better way.

michalszynkiewicz (approved):

PR #221, opened 2022-01-29
Implements a load-balancer following the "Power of Two Random Choices" strategy

### LINKED ISSUE #220: Provide a "Power of Two Random Choices" load-balancing strategy
See http://www.eecs.harvard.edu/~michaelm/NEWWORK/postscripts/twosurvey.pdf.

cescoffier:
Fix #220

Also:
- Use SecureRandom instead of Random
- Add a test in the static list discovery verifying that parsing errors are caught
- Add missing modules to the coverage
- Fix revapi package issues

### DIFF
--- a/load-balancer/power-of-two-choices/src/main/java/io/smallrye/stork/loadbalancer/poweroftwochoices/PowerOfTwoChoicesLoadBalancer.java
+++ b/load-balancer/power-of-two-choices/src/main/java/io/smallrye/stork/loadbalancer/poweroftwochoices/PowerOfTwoChoicesLoadBalancer.java
@@ -1,0 +1,51 @@
+package io.smallrye.stork.loadbalancer.poweroftwochoices;
+
+import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+
+import io.smallrye.stork.api.LoadBalancer;
+import io.smallrye.stork.api.NoServiceInstanceFoundException;
+import io.smallrye.stork.api.ServiceInstance;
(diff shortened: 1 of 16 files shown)


aureamunoz (approved):
LGTM

load-balancer/power-of-two-choices/src/main/java/io/smallrye/stork/loadbalancer/poweroftwochoices/PowerOfTwoChoicesLoadBalancer.java
```diff
@@ -0,0 +1,51 @@
...
+
+        List<ServiceInstance> instances = new ArrayList<>(serviceInstances);
+        int count = instances.size();
+
+        if (count == 1) {
+            return instances.get(0);
+        }
+
+        ServiceInstance first = instances.get(random.nextInt(count));
+        ServiceInstance second = instances.get(random.nextInt(count));
```
michalszynkiewicz, line 38: shouldn't it select a different random instance?

michalszynkiewicz: actually, we select these random two every time so the fact that we sometimes have the same two is not an issue

cescoffier: yes, it's not a problem.
I was wondering to do an object equality check in this case, but it's clearly premature optimization.

PR #222, opened 2022-01-29
Replace raw types from Metadata to use properly parameterized types

cescoffier:
It is a breaking change. However, the radius blast is minimal, if not invisible to the users (except fewer warnings). It will also prevent invalid access to the metadata. The revapi contains the changes and justifications.

This commit also adds a few tests to the API module. It also fixes the revapi configuration of the core module to avoid overlap with the API module.


aureamunoz (approved):
LGTM :-) THx

PR #238, opened 2022-02-15
Improved statistics collection for load balancers

michalszynkiewicz:


api/src/main/java/io/smallrye/stork/api/ServiceInstance.java
```diff
@@ -50,11 +50,27 @@ default boolean gatherStatistics() {
...
+     * When {@code gatherStatistics} is enabled, reports the failure of an operation using this service instance.
+     * <p>
+     * If the call for which the error is observed is guided by recordStart/recordEnd, this method should be called <b>after</b>
+     * {@link #recordEnd()}.
      *
-     * @param timeInNs the duration of the operation in nano-seconds.
      * @param failure the failure if the operation failed. Can be {@code null}.
      */
-    default void recordResult(long timeInNs, Throwable failure) {
+    default void recordError(Throwable failure) {
```
cescoffier, line 74: I would use "failure" instead of error. That's the wording we used in mutiny.

api/src/main/java/io/smallrye/stork/api/ServiceInstance.java
```diff
@@ -50,11 +50,27 @@ default boolean gatherStatistics() {
...
+     */
+    default void recordStart() {
+    }
+
+    /**
+     * When {@code gatherStatistics} is enabled, reports the end of an operation using this service instance.
+     *
+     * <b>Must</b> be called for each call for which {@link #recordStart()} was called, regardless of recording errors
+     */
+    default void recordEnd() {
```
cescoffier, line 63: I'm not sure to understand the model.
Basically, in case of success, I got end called. In case of failure, I got error and end called. So, when error is called, I need to store the exception and the processing I need to do to handle that error need to happen in the end method.

api/src/main/java/io/smallrye/stork/api/ServiceInstance.java
```diff
@@ -50,11 +50,27 @@ default boolean gatherStatistics() {
...
+     * If the call for which the error is observed is guided by recordStart/recordEnd, this method should be called <b>after</b>
+     * {@link #recordEnd()}.
      *
-     * @param timeInNs the duration of the operation in nano-seconds.
      * @param failure the failure if the operation failed. Can be {@code null}.
      */
-    default void recordResult(long timeInNs, Throwable failure) {
+    default void recordError(Throwable failure) {
     }
 }
```
cescoffier, line 76: I think I'm confused because it looks like a uni, but it's not a uni in the sense that in a uni only end or error would be called as both would be considered terminal. 

What's the rationale for such a model (I would try to avoid having another model)

michalszynkiewicz: For grpc we can sensibly measure times only for unary calls. But for streaming we can still collect errors, and should take them into account. Hence the separation

cescoffier: Aren't failure terminal operation in gRPC streams too?

michalszynkiewicz: Streaming calls are not wrapped in recordStart/recordEnd at the moment. We don't want to measure times for streaming connections. For load balancers that focus on inflight calls, this would make sense though.
We could change this API even further and have sth like this:
```
recordStart(); // invoked on a start of a call
recordReply(); // invoked on a response, this would be the moment to measure time
recordEnd(@Nullable Throwable failure); // on call end, for both unary and streams, this would be the moment to decrease inflight calls no
```

WDYT?

cescoffier: Yes, looks better. 
Is it a reply (as in request/reply) or a response?

michalszynkiewicz: it's a reply/response (ain't they synonyms?) for a request in case of a unary operation.
Actually, to make a proper use of it in load balancers that measure time it would have to be something like this, I think:
```java
recordStart(boolean measureTime); // invoked on a start of a call
recordReply(); // invoked on a response, this would be the moment to measure time
recordEnd(@Nullable Throwable failure); // on call end, for both unary and streams, this would be the moment to decrease inflight calls no
```

cescoffier (approved):

PR #243, opened 2022-02-17
Serializing calls to non-threadsafe load balancers

michalszynkiewicz:


api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -12,31 +14,75 @@
...
      * The selection looks for the service instances and select the one to use using the load balancer.
      *
+     * <b>Note:</b> this method doesn't record a start of an operation using this load balancer and does not
+     * synchronize load balancer invocations even if the load balancer is not thread safe
+     *
      * @return a Uni with a ServiceInstance, or with {@link NoServiceInstanceFoundException} if the load balancer failed to find
      *         a service instance capable of handling a call
      * @throws IllegalArgumentException if the current service does not use a load balancer
      */
     public Uni<ServiceInstance> selectServiceInstance() {
```
michalszynkiewicz, line 44: should this method be renamed?

aureamunoz (approved):

api/src/main/java/io/smallrye/stork/api/Service.java
michalszynkiewicz: Maybe, since it's the Service class and Service is kind of implied, it could be `selectInstance` and `selectInstanceAndRecordStart`?

aureamunoz: +1, I like it

aureamunoz (approved):
LGTM

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
```diff
@@ -22,4 +22,10 @@
      *         are deemed invalid for some reason
      */
     ServiceInstance selectServiceInstance(Collection<ServiceInstance> serviceInstances);
+
+    /**
+     *
+     * @return true if the selectServiceInstance method can safely be called in parallel
```
cescoffier, line 28:
```suggestion
     * @return true if the selectServiceInstance method can safely be called concurrently.
```

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
```diff
@@ -22,4 +22,10 @@
...
+     */
+    boolean isThreadSafe();
```
cescoffier, line 30: Not sure I like the name. Your implementation can be thread-safe, but still does not work as expected on concurrent calls.

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -12,39 +15,110 @@
...
     private final boolean secure;
 
-    public Service(String serviceName, Optional<LoadBalancer> loadBalancer, ServiceDiscovery serviceDiscovery, boolean secure) {
+    public Service(String serviceName, Optional<LoadBalancer> loadBalancer, ServiceDiscovery serviceDiscovery,
+            boolean secure, boolean threadSafeLoadBalancer) {
         this.loadBalancer = loadBalancer;
         this.serviceDiscovery = serviceDiscovery;
         this.serviceName = serviceName;
         this.secure = secure;
+        instanceSelectionLock = threadSafeLoadBalancer ? null : new ReentrantLock();
```
cescoffier, line 30:
```suggestion
       this.instanceSelectionLock = threadSafeLoadBalancer ? null : new ReentrantLock();
```

cescoffier, line 30: why reentrant?

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -12,39 +15,110 @@
...
      * Selects a service instance.
+     * <p>
+     * The selection looks for the service instances and select the one to use using the load balancer.
+     *
+     * <b>Note:</b> this method doesn't record a start of an operation using this load balancer and does not
+     * synchronize load balancer invocations even if the load balancer is not thread safe
      *
+     * @return a Uni with a ServiceInstance, or with {@link NoServiceInstanceFoundException} if the load balancer failed to find
+     *         a service instance capable of handling a call
+     * @throws IllegalArgumentException if the current service does not use a load balancer
```
cescoffier, line 43: Can this happen? It should use the round-robin LB by default.

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -12,39 +15,110 @@
...
+     * Select a ServiceInstance for this service from a collection and record a start of an operation using the instance
+     *
+     * Access to the underlying LoadBalancer method is serialized
+     *
+     * @param instances instances to choose from
+     * @param measureTime whether the operation for which the operation is chosen records time (will call
+     *        {@link ServiceInstance#recordReply()})
+     * @return the selected instance
+     *
+     *
```
cescoffier, line 92: Too many blank lines.

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
cescoffier: We should have a default implementation.

load-balancer/least-response-time/src/main/java/io/smallrye/stork/loadbalancer/leastresponsetime/LeastResponseTimeLoadBalancer.java
```diff
@@ -42,6 +46,10 @@ public ServiceInstance selectServiceInstance(Collection<ServiceInstance> service
         for (ServiceInstance instance : serviceInstances) {
             CallStatistics.CallsData callsData = callStatistics.statsForInstance(instance.getId());
             if (callsData == null || callsData.lastRecorded == CallStatistics.NO_CALL_STARTED) {
+                if (log.isDebugEnabled()) {
```
cescoffier, line 49: I don't think the check is required. log.debug will do it.

load-balancer/least-response-time/src/main/java/io/smallrye/stork/loadbalancer/leastresponsetime/LeastResponseTimeLoadBalancer.java
```diff
@@ -52,6 +60,9 @@ public ServiceInstance selectServiceInstance(Collection<ServiceInstance> service
                 if (score < bestScore) {
                     best = instance;
                     bestScore = score;
+                    if (log.isDebugEnabled()) {
```
cescoffier, line 63: Same here.

load-balancer/least-response-time/src/main/java/io/smallrye/stork/loadbalancer/leastresponsetime/LeastResponseTimeLoadBalancer.java
```diff
@@ -62,10 +73,23 @@ public ServiceInstance selectServiceInstance(Collection<ServiceInstance> service
                     ? ((List<ServiceInstance>) serviceInstances).get(selectedIdx)
                     : new ArrayList<>(serviceInstances).get(selectedIdx);
         }
+        if (log.isDebugEnabled()) {
```
cescoffier, line 76: Same here.

cescoffier (changes requested):

api/src/main/java/io/smallrye/stork/api/Service.java
michalszynkiewicz: true, I forgot about it

api/src/main/java/io/smallrye/stork/api/Service.java
michalszynkiewicz: That's the simplest thing java brings...
I think I'll change it to semaphore. The reentrant lock is bound to a thread, a single thread can enter it multiple times. I don't think we want it.

michalszynkiewicz: I don't like those thises :D but you're right it looks better if all assignments have them than with one not having it

api/src/main/java/io/smallrye/stork/api/Service.java
michalszynkiewicz: I'll change Optional<LoadBalancer> to LoadBalancer for the LB field type here too

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
michalszynkiewicz: I'll extend this javadoc

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
michalszynkiewicz: Maybe sth like this?
```java
default boolean requiresStrictStatCollection() {
   return true;
}
```

load-balancer/least-response-time/src/main/java/io/smallrye/stork/loadbalancer/leastresponsetime/LeastResponseTimeLoadBalancer.java
michalszynkiewicz: true, the ternarny operator is probably cheaper then the if

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
michalszynkiewicz: I change the javadoc entirely, WDYT of the current version?

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
cescoffier: Yes, looks better. 

`true` by default is the safest, so +1.

api/src/main/java/io/smallrye/stork/api/LoadBalancer.java
aureamunoz: +1 it's clearer now I think.

PR #247, opened 2022-02-21
'use-secure-random' attribute

cescoffier:


aureamunoz (changes requested):
The documentation would be missing, except that LGTM!

cescoffier:
The attribute is going to be part of the documentation table, so included in each load-balancer page. I don't believe we need a section about that parameter.

cescoffier:
Actually... no, the tables are not copied in the LB. Fixing...

cescoffier:
@aureamunoz fixed!

aureamunoz (approved):
LGTM!

PR #254, opened 2022-02-24
Deprecate the "secure" ServiceConfig attribute in favor to service discovery "secure" attributes.

cescoffier:
- Implement a backward-compatible way
- Add the "secure" attribute to all service discovery (with slightly different documentation)
- Add test to verify the previous behavior and new behavior
- Declare the log level stating that round-robin is used by default

### DIFF
--- a/service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryProvider.java
+++ b/service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryProvider.java
@@ -13,14 +13,14 @@ import io.vertx.core.Vertx;
 @ServiceDiscoveryAttribute(name = "use-health-checks", description = "Whether to use health check.", defaultValue = "true")
 @ServiceDiscoveryAttribute(name = "application", description = "The application name; if not defined Stork service name will be used.")
 @ServiceDiscoveryAttribute(name = "refresh-period", description = "Service discovery cache refresh period.", defaultValue = "5M")
+@ServiceDiscoveryAttribute(name = "secure", description = "whether the connection with the service should be encrypted with TLS.")
 @ServiceDiscoveryType("consul")
 public class ConsulServiceDiscoveryProvider implements ServiceDiscoveryProvider<ConsulServiceDiscoveryProviderConfiguration> {
 
     @Override
     public ServiceDiscovery createServiceDiscovery(ConsulServiceDiscoveryProviderConfiguration config, String serviceName,
             ServiceConfig serviceConfig, StorkInfrastructure storkInfrastructure) {
-        return new ConsulServiceDiscovery(serviceName, config, storkInfrastructure.get(Vertx.class, Vertx::vertx),
-                serviceConfig.secure());
+        return new ConsulServiceDiscovery(serviceName, config, storkInfrastructure.get(Vertx.class, Vertx::vertx));
 
     }
 }
--- a/service-discovery/static-list/revapi.json
+++ b/service-discovery/static-list/revapi.json
@@ -24,7 +24,24 @@
     "criticality" : "highlight",
     "minSeverity" : "POTENTIALLY_BREAKING",
     "minCriticality" : "documented",
-    "differences" : [ ]
+    "differences" : [
+      {
+        "ignore": true,
+        "code": "java.annotation.removed",
+        "old": "class io.smallrye.stork.servicediscovery.staticlist.StaticListServiceDiscoveryProvider",
+        "new": "class io.smallrye.stork.servicediscovery.staticlist.StaticListServiceDiscoveryProvider",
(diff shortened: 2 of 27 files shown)


aureamunoz (changes requested):

service-discovery/static-list/revapi.json
```diff
@@ -24,7 +24,24 @@
...
+        "justification": "By adding a second annotation, it switches to 'ServiceDiscoveryAttributes'. Should not impact users."
+      },
+      {
+        "ignore": true,
+        "code": "java.annotation.added",
+        "old": "class io.smallrye.stork.servicediscovery.staticlist.StaticListServiceDiscoveryProvider",
+        "new": "class io.smallrye.stork.servicediscovery.staticlist.StaticListServiceDiscoveryProvider",
+        "annotation": "@io.smallrye.stork.api.config.ServiceDiscoveryAttributes({@io.smallrye.stork.api.config.ServiceDiscoveryAttribute(name = \"address-list\", description = \"A comma-separated list of addresses (host:port). The default port is 80.\"), @io.smallrye.stork.api.config.ServiceDiscoveryAttribute(name = \"secure\", description = \"Whether the connection with the service should be encrypted with TLS. Default is false, except if the host:port uses the port is 443.\")})",
+        "justification": "By adding a second annotation, it switches to 'ServiceDiscoveryAttributes'. Should not impact users."
```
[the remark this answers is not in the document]
cescoffier: That's the beauty of Java...
Before we had a single annotation, so it was: `ServiceDiscoveryAttribute`.
Now, we have 2 annotations, so Java introduces a container annotation `ServiceDiscoveryAttributes` (plural), which contains the 2 `ServiceDiscoveryAttribute` (backed in an array). 

And yes, it can be seen as a breaking change for programs looking at the `ServiceDiscoveryAttribute` annotations, which would not find anything anymore as they need to look at `ServiceDiscoveryAttributes` and gets the array of `ServiceDiscoveryAttribute`. No worries for us, we already do that work.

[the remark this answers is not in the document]
aureamunoz: ok. I see yes, but the `address-list` did existe for a while now, right? It's just that the revapi hadn't traced it when it wass added, no?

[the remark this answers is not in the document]
cescoffier: `address-list` was the top level `@ ServiceDiscoveryAttribute`. Now it's in the nested array:

```
@ ServiceDiscoveryAttributes( {
   @ ServiceDiscoveryAttribute, // address-list
   @ ServiceDiscoveryAttribute // secure
})
```

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryProvider.java
```diff
@@ -13,14 +13,14 @@
 @ServiceDiscoveryAttribute(name = "use-health-checks", description = "Whether to use health check.", defaultValue = "true")
 @ServiceDiscoveryAttribute(name = "application", description = "The application name; if not defined Stork service name will be used.")
 @ServiceDiscoveryAttribute(name = "refresh-period", description = "Service discovery cache refresh period.", defaultValue = "5M")
+@ServiceDiscoveryAttribute(name = "secure", description = "whether the connection with the service should be encrypted with TLS.")
```
michalszynkiewicz, line 16: shouldn't we assume it's always there and just generate code (and docs) for it regardless of the anntation being here?

cescoffier: That was like that before , without the doc part.

The problem is that the semantic of this attribute depends on the protocol. For example, with Eureka it influences the endpoint selection. That's why it better to have it in each service discovery with an appropriate description.

michalszynkiewicz: > The problem is that the semantic of this attribute depends on the protocol. For example, with Eureka it influences the endpoint selection. That's why it better to have it in each service discovery with an appropriate description.

Do you mean that the documentation may be different?

michalszynkiewicz: do you think there are service discovery strategies that won't have this attribute?

cescoffier: in the future, definitely. With TLS becoming the norm, it is likely that some service discovery providers will require that all the services use TLS. In this case, these service discovery providers will produce service instances with `secure=true`, but would not have the attribute, because it would be always true. They may reject instances using plain text.

When will we see this? It's hard to say. The main issue is that TLS is often seen as black magic and convoluted to set up. Once these technical difficulties are removed, it will be TLS everywhere. 

image

PR #259, opened 2022-03-09
Stork Programmatic API proposal

cescoffier:
- Fix javadoc and identation of the generated config and service loader classes
- Generate with methods in the generated config classes
- Make generated config classes implement the ServiceDiscoveryConfig and LoadBalancerConfig interfaces
- Implement the programmatic API


michalszynkiewicz:
So a single config class after all?

I like it. I have some small doubts if we need to expose both `of` and `with...` methods in the `ServiceDefinition` class.

cescoffier:
> So a single config class after all?

Yes, but I made the class immutable. I started with two files, but it makes things more complicated.

> I like it. I have some small doubts if we need to expose both of and with... methods in the ServiceDefinition class.

I'm still not sure at the moment. I added `of` this morning to see what it looks like from the user POV. 


Once thing that needs to be done (but that's a breaking change) is renaming the generated class.

cescoffier:
@michalszynkiewicz renaming done....

Renaming: 3 minutes
Updating code: 10 minutes
Adding justifications fo revapi: 30 minutes :-D

michalszynkiewicz (approved):
LGTM, the only doubt, as I wrote previously, is if we should do both `of` and `with`. But I don't have a preference which is better :)

cescoffier:
@aureamunoz @vsevel do you prefer .with or .of ?

I know, this is an existential question.

Ladicek:
Assuming the `with` or `of` question concerns the `ServiceDefinition` class ... is there any reason whatsoever to keep that class mutable? Make it `final` and immutable and your problem is solved :-)

cescoffier:
I can make this class immutable.

cescoffier:
I have a version making `ServiceDefinition` immutable (so removing the `with` methods from `ServiceDefinition`). However, I've found a problem... the composite service discovery has an `initialize` callback which is called when stork is initialized which would not be working with this programmatic approach. I'm working on that right now.

cescoffier:
@michalszynkiewicz The last commit is a lot deeper. I had to change some APIS in order to work around the initialization issue.

aureamunoz (approved):
LGTM.
This should be also documented, right? In order to show how to use this new api. I imagine it's in progress or it will come later :-)

aureamunoz:
> @aureamunoz @vsevel do you prefer .with or .of ?
> 
> I know, this is an existential question.

I would prefer `with` in this case.

michalszynkiewicz:
@cescoffier the only thing I don't like is that we have <type>Configuration for service discovery and <type>LoadBalancerConfiguration for load balancers.
Would be more consistent if it was <type>ServiceDiscoveryConfiguration for the former.

cescoffier:
Hum, that's unexpected. Let me have a look. It should be consistent. I actually had to add "load-balancer" or "selector" to some tests as it was regenerating the same file (if you service discovery and selection use the same name).

cescoffier:
@michalszynkiewicz I think you got confused by the name conflict.
For example, here is a test I just wrote:
```java
stork
                .defineIfAbsent("first-service", ServiceDefinition.of(
                        new StaticConfiguration().withAddressList(listOfServices),
                        new PowerOfTwoChoicesConfiguration()))
```

Both configuration classes are named with `camel(type) + "Configuration"`. 


There are some tests where you see "$TypeLoadBalancerConfiguration", because I had to change the type to "$type-load-balancer". When you have a service discovery and a load balancer in the same package with the same type (in other words in my tests, as this is particularly odd), the annotation processor cannot create the load balancer configuration class. The reason is simple: the class already exists. It was created for the service discovery ($type + "Configuration").

michalszynkiewicz:
Got it now. Thanks for the explanation

michalszynkiewicz (approved):

michalszynkiewicz:
Do you want to squash the commits?

cescoffier:
I'm not done yet :-)
Need a bit more tests and doc. Will squash when I have everything.

cescoffier:
@michalszynkiewicz @aureamunoz Ok, doc and tests are there.

cescoffier:
Actually, there is no need to squash the commits. They are all "steps". 
Now, I can do it, just let me know.

api/src/main/java/io/smallrye/stork/api/Stork.java
```diff
@@ -0,0 +1,47 @@
+package io.smallrye.stork.api;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * The central API of Stork
+ */
+public interface Stork {
```
michalszynkiewicz, line 9: Won't having interface named the same as the central entrypoint class reduce discoverability?

michalszynkiewicz: Maybe the interface could be called `StorkServiceRegistry`?

api/src/main/java/io/smallrye/stork/api/ServiceDiscovery.java
```diff
@@ -22,9 +21,8 @@
      * Optional initialization.
      * This method will be invoked after all service discoveries and load balancers are registered in Stork
      *
-     * @param services a map of all defined services by name, helpful with creating service discovery implementations
-     *        that combine results from multiple services
+     * @param stork the stork instance managing the service.
      */
-    default void initialize(Map<String, Service> services) {
+    default void initialize(Stork stork) {
```
michalszynkiewicz, line 26: was this change required?

cescoffier: Actually, yes. By doing that, service discovery providers can register new services after the initialization.

api/src/main/java/io/smallrye/stork/api/Stork.java
cescoffier: I agree the name is not great, and it's confusing.

The problem is that we are leaking the Stork from the core API (because of the `initialize` method) and we also have an API module, which does not contain the main API... 

`StorkServiceRegistry` would work as name. I will rename it.

cescoffier:
Renamed the Stork class and squashed the commit

michalszynkiewicz (approved):

PR #270, opened 2022-03-24
Document static list service discovery

### LINKED ISSUE #239: Missing documentation fo the "static" service discovery
When relying on DNS resolution, it can be useful. So, it should be documented.

michalszynkiewicz:
fixes #239


docs/service-discovery/static-list.md
```diff
@@ -0,0 +1,29 @@
...
+    <version>{{version.current}}</version>
+</dependency>
+```
+
+## Configuration
+
+For each service that should use the static list of service instances configure the service discovery `type`:
+
+```properties
+stork.my-service.service-discovery.type=static
```
cescoffier, line 23: You should mention the address-list, as it's mandatory.

cescoffier (changes requested):

aureamunoz (approved):
LGTM

docs/service-discovery/static-list.md
michalszynkiewicz: This is how this page is rendered, I didn't want to duplicated the info on it being mandatory but I'm fine adding it
image

cescoffier (approved):

cescoffier (approved):

PR #274, opened 2022-03-28
`sticky` LoadBalancer implementation

### LINKED ISSUE #271: sticky load balancer
Select one service instance and use it until it fails. Then switch to a different one.

michalszynkiewicz:
fixes #271

### DIFF
--- a/docs/load-balancer/sticky.md
+++ b/docs/load-balancer/sticky.md
@@ -1,0 +1,30 @@
+# Sticky Load Balancing
+
+The sticky load balancer selects a single service instance and keeps using it until it fails.
+Then, it selects another one.
+
+It is possible to configure a backoff time to specify for how long a failing service instance should not be retried.
+
+Precisely, the load balancer works as follows:
+
+* if no service instance has been selected so far, select the first instance from the collection;
+* else if the previously selected service instance has not failed, and is still available, return it;
+* else return the first available service instance that has no recorded failure, if one exists;
+* else, find the available instance for which the time since the last failure is the longest, and
+    * if the backoff time since the failure passed, return it;
+    * or, throw an `NoAcceptableServiceInstanceFoundException` as no acceptable instances are available.
+
+## Configuration
+
+To use the `sticky` load service selection strategy, set the load balancer type to `sticky`:
+
+```properties
+stork.my-service.service-discovery=...
+stork.my-service.service-discovery...=...
+stork.my-service.load-balancer.type=sticky
+```
+
+
+The following attributes are supported:
+
+--8<-- "load-balancer/sticky/target/classes/META-INF/stork-docs/sticky-lb-attributes.txt"
--- a/load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancer.java
+++ b/load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancer.java
@@ -1,0 +1,110 @@
+package io.smallrye.stork.loadbalancer.random;
+
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import io.smallrye.stork.api.LoadBalancer;
+import io.smallrye.stork.api.NoAcceptableServiceInstanceFoundException;
+import io.smallrye.stork.api.NoServiceInstanceFoundException;
+import io.smallrye.stork.api.ServiceInstance;
+import io.smallrye.stork.impl.ServiceInstanceWithStatGathering;
+import io.smallrye.stork.spi.CallStatisticsCollector;
+
+/**
+ * Select a single instance and use it until it fails.
+ * On failure, store the time of failure in `failedInstances` map.
+ * <p/>
+ * When new instance selection is needed:
+ * <ul>
+ * <li>if there's an instance for which we don't have a failure recorded, use it</li>
+ * <li>otherwise, pick the instance whose failure was the longest time away and:
+ * <ul>
+ * <li>return it if {@code failureBackOff} has passed since its last failure</li>
+ * <li>throw NoAcceptableServiceInstanceFoundException if it has not</li>
+ * </ul>
+ * </li>
+ * </ul>
+ */
+public class StickyLoadBalancer implements LoadBalancer, CallStatisticsCollector {
+
+    private final long failureBackOffNs;
+
+    private final LinkedHashMap<Long, Long> failedInstances = new LinkedHashMap<>();
+    private volatile ServiceInstanceWithStatGathering lastSelected;
+
+    public StickyLoadBalancer(Duration failureBackOff) {
+        this.failureBackOffNs = failureBackOff.toNanos();
+    }
+
+    // TODO start randomly, maybe sort the collection by service instance ids
+    // TODO: flow chart for how it works alongside the text
+    @Override
+    public ServiceInstance selectServiceInstance(Collection<ServiceInstance> serviceInstances) {
+        if (serviceInstances.isEmpty()) {
+            throw new NoServiceInstanceFoundException("No service instance found");
+        }
+
+        if (lastSelected != null) {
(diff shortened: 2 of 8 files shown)


docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
+# Sticky Load Balancing
+
+The sticky load balancer selects a single service instance and keeps using it until it fails.
+Then, it selects the next one.
+
+It is possible to configure a back off time to specify for how long a service instance that failed should not be retried.
```
cescoffier, line 6:
```suggestion
It is possible to configure a backoff time to specify for how long a failing service instance should not be retried.
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+
+Precisely, the load balancer works as follows:
+
+* if the previously selected service instance has not failed, and is still available, return it
```
cescoffier, line 10:
```suggestion
* if the previously selected service instance has not failed, and is still available, return it;
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+* else if exists, return the first available service instance that has no recorded failure
```
cescoffier, line 11:
```suggestion
* else if the service instance is still available, but has failed during a previous invocation, return the first available service instance that has no recorded failure;
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+* else, find the available instance that has failed the longest time ago, and
```
cescoffier, line 12:
```suggestion
* otherwise, find the available instance that was the first to record a failure, and
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+    * if the back-off time since failure passed, return it
```
cescoffier, line 13:
```suggestion
    * if the backoff time since the failure passed, return it;
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+    * else, report lack of applicable services (throw `NoAcceptableServiceInstanceFoundException`)
```
cescoffier, line 14:
```suggestion
    * or, throw an `NoAcceptableServiceInstanceFoundException` as no acceptable instances are available.
```

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+```properties
+stork.my-service.service-discovery=...
+stork.my-service.service-discovery...=...
+stork.my-service.load-balancer.type=sticky
+```
+
+
+The following attributes are supported:
+
+--8<-- "load-balancer/least-response-time/target/classes/META-INF/stork-docs/least-response-time-lb-attributes.txt"
```
cescoffier, line 29: That looks wrong...

load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancer.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+            failedInstanceIterator.remove();
+            ServiceInstance serviceInstance = serviceInstances.get(instanceId);
+            if (serviceInstance != null) {
+                lastSelected = new ServiceInstanceWithStatGathering(serviceInstance, this);
+                return lastSelected;
+            }
+        }
+        throw new NoAcceptableServiceInstanceFoundException("Each of the available service instances failed " +
+                "within the configured failure-back-off-time");
```
cescoffier, line 88: I think we should use `backoff`

load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancer.java
```diff
@@ -0,0 +1,108 @@
...
+    void recordEndAtTime(long serviceInstanceId, Throwable error, long currentTime) {
+        if (error != null) {
+            failedInstances.put(serviceInstanceId, currentTime);
+            lastSelected = null;
+        }
+    }
+
+    @Deprecated // for tests only
+    LinkedHashMap<Long, Long> getFailedInstances() {
+        return failedInstances;
```
cescoffier, line 106: to avoid misusage, should we return a copy?

load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancerProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+@LoadBalancerType(StickyLoadBalancerProvider.TYPE)
+@LoadBalancerAttribute(name = StickyLoadBalancerProvider.FAILURE_BACK_OFF_TIME, defaultValue = "0", description = "After how much time, "
+        +
+        "a service instance that has failed can be reused")
+public class StickyLoadBalancerProvider
+        implements LoadBalancerProvider<StickyConfiguration> {
+
+    static final String TYPE = "sticky";
+    static final String FAILURE_BACK_OFF_TIME = "failure-back-off-time";
```
cescoffier, line 18: I think we should use `backoff`

cescoffier (changes requested):

docs/load-balancer/sticky.md
michalszynkiewicz: the corrected sentences is not true. If no service was selected previously, we should fall under this branch too. Also it may happen that the previously selected instance did not fail but is just not availalbe anymore (service discovery-wise)

docs/load-balancer/sticky.md
michalszynkiewicz: that's not technically true. 
Let's assume we have services A and B.
A failed first, then B failed while backoff time had passed for A. So we used A. Then A failed.
At this point we don't select A.

michalszynkiewicz: I reworded it a bit

load-balancer/sticky/src/main/java/io/smallrye/stork/loadbalancer/random/StickyLoadBalancer.java
michalszynkiewicz: done

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
+# Sticky Load Balancing
+
+The sticky load balancer selects a single service instance and keeps using it until it fails.
+Then, it selects the next one.
```
cescoffier, line 4: instead of `next`, I would say `another` one.

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+
+It is possible to configure a backoff time to specify for how long a failing service instance should not be retried.
+
+Precisely, the load balancer works as follows:
+
+* if the previously selected service instance has not failed, and is still available, return it;
```
cescoffier, line 10: Maybe add the initial state: if no instance has been selected, select the first instance from the list.
(actually, random may be good too).

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+* else if exists, return the first available service instance that has no recorded failure;
```
cescoffier, line 11: I would reword the if exists: if the instance is still available, but has failed a previous invocation...

docs/load-balancer/sticky.md
```diff
@@ -0,0 +1,29 @@
...
+* else, find the available instance for which the time since the last failure is the greatest, and
```
cescoffier, line 12:
```suggestion
* else, find an available instance for which the time since the last failure is the longest, and
```

cescoffier (changes requested):

docs/load-balancer/sticky.md
michalszynkiewicz: Why `an` and not `the`? We're talking about a specific instance here.

cescoffier (approved):

PR #279, opened 2022-03-30
Improve javadoc and reduce some API visibility

cescoffier:
This commit prevents warnings during the javadoc generation.


michalszynkiewicz:
thanks @cescoffier !

PR #298, opened 2022-04-29
Service Registration SPI 

aureamunoz:


service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrar.java
```diff
@@ -44,7 +45,7 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<Kubernetes
             future.complete();
         }, result -> {
             if (result.succeeded()) {
-                log.error("Instances of service {} has been resgistered ", serviceName);
+                log.info("Instances of service {} has been resgistered ", serviceName);
```
cescoffier, line 48:
```suggestion
                log.info("Instances of service {} has been registered ", serviceName);
```

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrar.java
```diff
@@ -80,21 +81,31 @@ private void registerKubernetesService(String application, Metadata<KubernetesMe
                     application + "-" + ipAsSuffix(ipAddress), namespace, null, UUID.randomUUID().toString());
             EndpointAddress endpointAddress = new EndpointAddressBuilder().withIp(ipAddress).withTargetRef(targetRef)
                     .build();
-            Endpoints endpoint = new EndpointsBuilder()
-                    .withNewMetadata().withName(application).withLabels(serviceLabels).endMetadata()
-                    .addToSubsets(new EndpointSubsetBuilder().withAddresses(endpointAddress)
-                            .addToPorts(new EndpointPortBuilder().withPort(8080).build())
-                            .build())
+            EndpointSubset endpointSubset = new EndpointSubsetBuilder().withAddresses(endpointAddress)
+                    .addToPorts(new EndpointPortBuilder().withPort(8080).build())
```
cescoffier, line 85: I'm wondering about the hardcoded 8080 port. Shouldn't we look at the service object to extract the port?

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrationTest.java
```diff
@@ -43,7 +42,7 @@ void setUp() {
     }
 
     @Test
-    @Disabled("doesn't work yet")
+//    @Disabled("doesn't work yet")
```
cescoffier, line 45: Should we remove the line completely?

cescoffier (changes requested):

cescoffier:
@aureamunoz there is a formatting issue. Can you re-apply the formatted and squash/push?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrar.java
aureamunoz: Yes, the hardcoding is not good. We need to pass the port instead.

aureamunoz:
I did some modifs according to the review, rebases onto main and finally changed the base branch of the PR.
The revapi is broken...

api/src/main/java/io/smallrye/stork/api/Metadata.java
```diff
@@ -119,4 +118,12 @@ public String getName() {
         }
     }
 
+    public enum NoMetadataKey implements MetadataKey {
+        ;
```
cescoffier, line 122: why the semi-colon?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrarProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.smallrye.stork.api.ServiceRegistrar;
+import io.smallrye.stork.api.config.ServiceRegistrarAttribute;
+import io.smallrye.stork.api.config.ServiceRegistrarType;
+import io.smallrye.stork.spi.ServiceRegistrarProvider;
+import io.smallrye.stork.spi.StorkInfrastructure;
+
+@ServiceRegistrarAttribute(name = "k8s-host", description = "The Kubernetes API host.")
+@ServiceRegistrarType(value = "kubernetes", metadataKey = KubernetesMetadataKey.class)
+public class KubernetesServiceRegistrarProvider
```
cescoffier, line 14: Do we need a service registrar for Kube? I would have guessed that the application is automatically registered and nothing need to be done from our side.

cescoffier (commented):
Looks good.
Documentation would be important as it's hard to see how this is going to be used.

The registration / unregistration happens on the service implementation side (or a 3rd tier entity), so it should be carefully documented.

api/src/main/java/io/smallrye/stork/api/Metadata.java
michalszynkiewicz: Isn't it needed between constants/values declaration and methods?

aureamunoz: yes, it's needed, otherwise it doesn't compile...

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceRegistrarProvider.java
aureamunoz: I have also this remark, shouldn't we register services depending on the service discovery type configured? wdyt @michalszynkiewicz ?

michalszynkiewicz: I think what Clement says is that for Kube, service registration has limited sense.

michalszynkiewicz: re having the same configuration for service registration and for service discovery, that's definitely a good idea

aureamunoz:
I converted to draft because realised that the doc is lacking. I need also to fix an eureka test that is failing

cescoffier (approved):

PR #341, opened 2022-07-21
Wait for Consul to be ready before starting the tests

cescoffier:


service-discovery/dns/src/test/java/io/smallrye/stork/servicediscovery/dns/DnsServiceDiscoveryTest.java
```diff
@@ -66,7 +67,8 @@ public class DnsServiceDiscoveryTest {
                 portBindings.bind(TCP_8500, Ports.Binding.empty());
                 hostConfig.withPortBindings(portBindings);
                 cmd.withHostConfig(hostConfig);
-            });
+            })
+            .waitingFor(Wait.forListeningPort());
```
michalszynkiewicz, line 71: do we still need [`await().atMost(10, TimeUnit.SECONDS).until(this::clientCanTalkToConsul);`](https://github.com/smallrye/smallrye-stork/pull/341/files#diff-f1c9f665355ac0e2a7396a7e1cb741d1a158cafc44cd4dc3f71c6eb5c6472b30R91) with this?

michalszynkiewicz (approved):

service-discovery/dns/src/test/java/io/smallrye/stork/servicediscovery/dns/DnsServiceDiscoveryTest.java
cescoffier: Unfortunately yes.
My fix avoid an npe when retrieving the mapper ports.

PR #389, opened 2022-10-14
Fix missing attribute tables and footer image in doc

cescoffier:
- Configure the footer overrides
- Fix paths to attribute tables


aureamunoz (approved):
LGTM

PR #436, opened 2022-12-13
Stork 2.x 

cescoffier:
- Bump project version to 2.x
- Switch to the Jakarta API
- Allows instantiating service discovery provider, load balancer provider and service register provider as CDI beans 

_TODO:_

* [x] switch the provided providers to CDI
* [x] document the new approach to implementing providers, indicating that the previous approach still works
* [x] add a parameter to avoid the SPI registration (explicit opt-in to keep the compatibility)
* [x] update various dependencies to their Jakarta version


cescoffier:
Manually merged.

PR #445, opened 2022-12-28
2.x auri

aureamunoz:


cescoffier:
There is a problem with Eureka. The latest spring dependencies are compiled for Java 17. We need a was to support JAva 11 too. 
If it's problematic, let's change our strategy to test eureka. I'm not sure how, but starting Eureka in the same JVM as the tests seems to not be easy in the future (even currently).

cescoffier:
@aureamunoz there is still the spring issue. These spring dependencies are java 17 only, and so cannot be used here. We need to rollback to versions that are compatible with java 11.

cescoffier (approved):

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     strategy:
       matrix:
         java: [
-          { 'version': '11', opts: '' },
+          { 'version': '11', opts: '-pl ''!service-discovery/eureka''' },
```
cescoffier, line 52: you also need to skip the docs too, as it depends on that artifacts.

cescoffier (changes requested):

cescoffier:
Can you also reorganize the commits?

.github/workflows/build.yml
aureamunoz: yes, I've just seen that. But I don't understand why it fails on windows and not in ubuntu ¿?¿?

cescoffier: Pretty sure it's a question of cache.

cescoffier (approved):
Let's see what the CI says.

aureamunoz:
I can't believe it :sweat_smile:
Ups, I thought having chosen the 2.x branch as base, how do we proceed?

cescoffier:
Let me merge it manually tomorrow. I need to create a 1.x maintenance branch first.

cescoffier:
Manually merged.

PR #547, opened 2023-05-05
Adding new property `port-name` in K8s to select a port by name

Sgitario:
This is especially necessary when having endpoints that expose multiple ports.  At the moment, when having multiple ports, Stork does not select any and uses port 80 which is problematic when the application does not use this port. 

Relates https://github.com/quarkusio/quarkus/issues/33131

### DIFF
--- a/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
+++ b/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
@@ -18,6 +18,7 @@ import io.vertx.core.Vertx;
 @ServiceDiscoveryAttribute(name = "k8s-host", description = "The Kubernetes API host.")
 @ServiceDiscoveryAttribute(name = "k8s-namespace", description = "The namespace of the service. Use all to discover all namespaces.")
 @ServiceDiscoveryAttribute(name = "application", description = "The Kubernetes application Id; if not defined Stork service name will be used.")
+@ServiceDiscoveryAttribute(name = "port-name", description = "The Kubernetes application port name. If not defined, when exposing multiple ports, Stork will use the first one.")
 @ServiceDiscoveryAttribute(name = "refresh-period", description = "Service discovery cache refresh period.", defaultValue = CachingServiceDiscovery.DEFAULT_REFRESH_INTERVAL)
 @ServiceDiscoveryAttribute(name = "secure", description = "Whether the connection with the service should be encrypted with TLS.")
 @ApplicationScoped
--- a/service-discovery/kubernetes/revapi.json
+++ b/service-discovery/kubernetes/revapi.json
@@ -25,7 +25,17 @@
     "minSeverity" : "POTENTIALLY_BREAKING",
     "minCriticality" : "documented",
     "differences" : [
-
+      {
+        "ignore": true,
+        "code": "java.annotation.attributeValueChanged",
+        "old": "class io.smallrye.stork.servicediscovery.kubernetes.KubernetesServiceDiscoveryProvider",
+        "new": "class io.smallrye.stork.servicediscovery.kubernetes.KubernetesServiceDiscoveryProvider",
+        "annotationType": "io.smallrye.stork.api.config.ServiceDiscoveryAttributes",
+        "attribute": "value",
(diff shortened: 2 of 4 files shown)


Sgitario:
cc @aureamunoz

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
```diff
@@ -18,6 +18,7 @@
 @ServiceDiscoveryAttribute(name = "k8s-host", description = "The Kubernetes API host.")
 @ServiceDiscoveryAttribute(name = "k8s-namespace", description = "The namespace of the service. Use all to discover all namespaces.")
 @ServiceDiscoveryAttribute(name = "application", description = "The Kubernetes application Id; if not defined Stork service name will be used.")
+@ServiceDiscoveryAttribute(name = "port-name", description = "The Kubernetes application port name; if not defined and there is only one port defined, Stork will use it; otherwise it will use the port 80.")
```
aureamunoz, line 21: Where in the code are we taking the 80 port?

aureamunoz: I would refactor the 'otherwise' part of this to explain that this param is used to select the port when multiple ones are exposed, this is important and is not being mentioned.

aureamunoz (changes requested):
The solution looks great to me. Just a proposition to improve it:
- Could you please add a verification in one of the existing tests to check the case where there is only one exposed port (with and without `portname`configuration).

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
Sgitario: > Where in the code are we taking the 80 port?

I guess using the fallback service discovery: https://github.com/smallrye/smallrye-stork/blob/72b13e399293ee971079be9f8212807794cf1400/service-discovery/static-list/src/main/java/io/smallrye/stork/servicediscovery/staticlist/StaticListServiceDiscoveryProvider.java#L42, but I can't tell for sure.

> I would refactor the 'otherwise' part of this to explain that this param is used to select the port when multiple ones are exposed, this is important and is not being mentioned.

Will update it.

Sgitario:
PR updated with a new description of the parameter, let me know if it looks better now.

aureamunoz:
> PR updated with a new description of the parameter, let me know if it looks better now.

Yes, better now but still need to be clair about the 80 port (look my comment above). Also if you can add an assertion about the port in cases where we only have one port exposed ;-)

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
Sgitario: Yes, I know it's the static service discovery, though it won't use the container port, but 80 (I can't tell where in the code, 80 is selected, I haven't played much with stork). What is happening is that the port is 0 and somehow then it uses 80.

Sgitario:
> > PR updated with a new description of the parameter, let me know if it looks better now.
> 
> Yes, better now but still need to be clair about the 80 port (look my comment above). Also if you can add an assertion about the port in cases where we only have one port exposed ;-)

When there is only one port, there are test cases covering this scenario: https://github.com/smallrye/smallrye-stork/blob/83ac9c9b193cd2485eac7d6da9d9aa5bf37556a5/service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryTest.java#L86

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
aureamunoz: Mmm :thinking:. Given that we don't see clearly from where comes this 80 port, could you please say that container port will be take by default?

aureamunoz (changes requested):

Sgitario:
PR updated: Now, when having multiple ports, it will select the first one if the `port-name` is not provided.

PR #653, opened 2023-09-22
Expose selected instance

aureamunoz:


cescoffier:
@aureamunoz can you rebase from `main`? Seems like you have many commits.

aureamunoz:
> @aureamunoz can you rebase from `main`? Seems like you have many commits.

I did it and resolve a few conflicts. I will squash them

api/src/main/java/io/smallrye/stork/api/observability/ObservationPoints.java
```diff
@@ -115,6 +115,10 @@ public int getDiscoveredInstancesCount() {
         public Throwable failure() {
             return failure;
         }
+
+      public long getSelectedInstance() {
```
cescoffier, line 119: Maybe getSelectedInstanceId ? Otherwise I would have expected the ServiceInstance.

aureamunoz:
I think now is ok. I needed to rebase onto main also observability branch. Are you ok with it @cescoffier ?

PR #907, opened 2024-07-26
Add a few logs for registrars

aureamunoz:


service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -76,11 +76,14 @@ private Uni<Void> registerApplicationInstance(WebClient client, String applicati
                         .put("name", "MyOwn"))
                 .put("leaseInfo", new JsonObject().put("renewalIntervalInSecs", 10000).put("durationInSecs", 10000));
 
-        return client.post(path + "/eureka/apps/" + applicationId)
+        Uni<Void> response = client.post(path + "/eureka/apps/" + applicationId)
                 .putHeader("content-type", "application/json")
                 .putHeader("accept", "application/json")
                 .sendJson(instance).replaceWithVoid();
 
+        log.info("Instance registered for service {}: {}", applicationId, registration);
```
cescoffier, line 84: Should you add this log when the Uni is avec running? 

Before the .replaceWithVoid, use .invoke(() -> log.info....)

service-registration/static-list/src/main/java/io/smallrye/stork/serviceregistration/staticlist/StaticListServiceRegistrar.java
```diff
@@ -28,6 +28,7 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<Metadata.D
                 "service '" + serviceName + "'");
         String hostAndPortToAddString = StorkAddressUtils.parseToString(hostAndPortToAdd);
         InMemoryAddressesBackend.add(serviceName, hostAndPortToAddString);
+        log.info("Address %s has been registered for service %s", hostAndPortToAddString, serviceName);
```
cescoffier: Here it's not a problem, because it's synchronous.

cescoffier (changes requested):

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
aureamunoz: yes!

cescoffier (approved):

PR #940, opened 2024-09-18
Fix attributes path, include all deps in docs and some refactors

aureamunoz:
Oops, looks like there were a couple of things to fix


cescoffier (approved):
Awesome!

PR #956, opened 2024-10-11
Add clarification about how composite service discovery works.

aureamunoz:


docs/docs/service-discovery/composite.md
```diff
@@ -38,9 +38,14 @@ For each service that should consist of multiple services, configure the service
     quarkus.stork.serviceB.service-discovery.type=...
     ```
 
-
 Remember to define the services that make up your composite service.
 
+Be aware that Stork doesn't work as a standalone service discovery cluster. 
+Instead, it processes composite configurations—meaning it handles multiple configurations that may use different service discovery implementation. 
+For each specific configuration, Stork delegates the service discovery task to the appropriate service discovery provider. 
+If one of these providers doesn't respond or fails, it affects Stork's ability to resolve that specific configuration, but it doesn't mean Stork itself is faulty;
```
cescoffier, line 46: That's the current implementation, but shouldn't we only fail if none of the sub-discovery found a matching instance or if all fail?

Basically, if one fails but if another one provides a service instance, it should pass this service instance.

cescoffier (approved):
LGTM, added a comment because I believe we can improve the composite service discovery.

docs/docs/service-discovery/composite.md
aureamunoz: Yes, we can better manage the error case.
Do we do it in this PR? or shall we create an issue and do it in another one?

cescoffier:
In another one.

PR #959, opened 2024-10-16
Fix issue about empty consul instance host

aureamunoz:


service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscovery.java
```diff
@@ -74,8 +74,8 @@ private List<ServiceInstance> toStorkServiceInstances(ServiceEntryList serviceEn
             Metadata<ConsulMetadataKey> consulMetadata = createConsulMetadata(serviceEntry);
             String address = service.getAddress();
             int port = serviceEntry.getService().getPort();
-            if (address == null) {
-                throw new IllegalArgumentException("Got null address for service " + serviceName);
+            if (address == null || address.isEmpty() || address.isBlank()) {
+                address = serviceEntry.getNode().getAddress();
```
cescoffier, line 78: Can you mention in a comment why we do this choice. It should also be documented.

cescoffier (approved):

PR #1002, opened 2025-02-11
Configure our own annotation processor from config-generator module.

aureamunoz:
smallrye-parent introduced a specific maven-compiler-plugin configuration. So now we need to explicitly configure the plugin in Stork to use our annotation processor from the config-generator module.


cescoffier (approved):
@ogunalp can you have a look?

ozangunalp (approved):
LGTM!

PR #1019, opened 2025-03-31
Take scheme into account when deciding default port

geoand:
Fixes: https://github.com/quarkusio/quarkus/issues/47077


cescoffier (approved):

aureamunoz (changes requested):
I think we should do the same for [DNS](https://github.com/smallrye/smallrye-stork/blob/cb972b6019a510c7dd9ee687955e867f6124f037/service-discovery/dns/src/main/java/io/smallrye/stork/servicediscovery/dns/DnsServiceDiscovery.java#L71) using 853 in that case

cescoffier:
What would it mean for DNS? DoT is uncommon (DNS over TLS), and I'm not sure it would work anyway. I know that there are a few improvements in Netty to handle that.

aureamunoz:
Oh yeah, ok... I just thought it was good to do it because there is also that option. But I was still going to propose to do it in another PR because I have to take a closer look at it. So we can merge this one.

aureamunoz (approved):

geoand:
🙏🏽

cescoffier:
Please go ahead. I’m traveling tomorrow

PR #1027, opened 2025-04-11
Unable retries in k8s client and add test

### LINKED ISSUE #1021: KubernetesServiceDiscovery  is calling the fetchNewServiceInstances method multiple times
KubernetesServiceDiscovery is calling the fetchNewServiceInstances method multiple times after an invalidate event. This can lead to frequent hits on the Kubernetes API, causing slower responses in the application’s REST client.

Old refreshing logic in CachingServiceDiscovery calls fetchNewServiceInstances only once.
```java
uni.memoize().atLeast(this.refreshPeriod);
```

aureamunoz:
Fixes #1021 
This solution addresses the issue of unnecessary retries triggered by the default Kubernetes client configuration. The core change is to disable automatic retries, ensuring that the cluster is only contacted when necessary.

Additionally, a test has been added to validate that the cluster is contacted only when the cache is invalidated. The cache is invalidated only when the instance retrieval is successful. If an error occurs during instance retrieval, the cache will not be reset, as resetting it would lead to potential inconsistencies. This behavior is expected and correct because we want to avoid resetting the cache on failure to prevent inconsistent states.

With this solution, the system avoids creating a loop of requests to the Kubernetes cluster and ensures it is only queried when the user calls service.getInstances() with an invalidated cache.

This is a draft because I still need certainly to expose the retries config.

cc @cescoffier @balif

### DIFF
--- a/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
+++ b/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
@@ -83,6 +83,7 @@ public class KubernetesServiceDiscovery extends CachingServiceDiscovery {
 
         Config k8sConfig = new ConfigBuilder(base)
                 .withMasterUrl(masterUrl)
+                .withRequestRetryBackoffLimit(0) //TODO: expose this in the config
                 .withNamespace(namespace).build();
         this.client = new KubernetesClientBuilder().withConfig(k8sConfig).build();
         this.vertx = vertx;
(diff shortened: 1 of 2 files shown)


service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -83,6 +83,7 @@ public KubernetesServiceDiscovery(String serviceName, KubernetesConfiguration co
 
         Config k8sConfig = new ConfigBuilder(base)
                 .withMasterUrl(masterUrl)
+                .withRequestRetryBackoffLimit(0) //TODO: expose this in the config
```
cescoffier, line 86: I'm still not sure this is disabling the retries

cescoffier (changes requested):
Are you sure that it disables the retries? For me it disables the delay between retries.

aureamunoz:
cc @manusa

cescoffier (approved):

PR #1037, opened 2025-04-24
Temp 2.6.2

aureamunoz:


core/src/main/java/io/smallrye/stork/Stork.java
```diff
@@ -264,7 +264,16 @@ public static Stork getInstance() {
      * Closes the stork instance.
      */
     public static void shutdown() {
-        REFERENCE.set(null);
+        var previous = REFERENCE.getAndSet(null);
+        if (previous != null) {
+            previous.clear();
+        }
+        //        InMemoryAddressesBackend.clearAll();
```
cescoffier, line 271: I'm wondering why I kept that code commented. Honestly, I don't remember. Maybe something to cleanup.

cescoffier (approved):

Issue #299, opened 2022-05-04
External cache invalidation API


michalszynkiewicz:

cescoffier:
We have 3 issues related to caching:

- #299 (this one)
- #61 
- #284 

It would be great to post an update on each - pretty sure that some of them have been fixed. 
If not, it would be great if we could make some progress.

\CC @aureamunoz

PR #1093, opened 2025-09-12
Expose cache invalidation

aureamunoz:
Related to https://github.com/smallrye/smallrye-stork/issues/299

It also includes a couple of refactors in some tests. I realised that the tests are not entirely consistent across all service discovery implementations, but I didn't have the courage to do it in all of them...

Even now, I don't think it's worth duplicating the test classes to check the programmatic API. I could have done it in a test method, as is the case with K8s sd, but since I did it for a few classes, I left it as it is...


cescoffier (approved):
LGTM! Thanks!

PR #1105, opened 2025-10-01
Remove check about service discovery type null value

aureamunoz:
Since registration became a standalone feature, service discovery configuration is not mandatory anymore.


cescoffier:
Should we have a test?

cescoffier (approved):

PR #1118, opened 2025-10-30
Fix package name of sticky load-balancer

gsmet:
This will avoid a split package warning when using both random and sticky.

(Noticed while working on the `integration-tests/main` module in Quarkus which uses both...)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
         <version.smallrye-config>3.14.1</version.smallrye-config>
 
         <vertx.version>4.5.10</vertx.version>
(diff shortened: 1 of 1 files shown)


aureamunoz (approved):
LGTM

aureamunoz:
Thank you!

cescoffier (dismissed):

aureamunoz (changes requested):
Could you add the revapi justif to fix check compatibility job?

gsmet:
Could you point me to an example?

PR #1125, opened 2025-11-14
Fix incorrect key in CallStatistics.storeTime

songsong821:
This change replaces the callIdx with serviceInstanceId when putting new CallsData in the storage.


aureamunoz:
> I'll work on adding tests to validate these changes. Since I'm not very familiar with Java and need to set up the proper environment, it might take me a little while to get this done properly. I'll update the MR once the tests are ready.

Ok, no rush. Thank you. I can give some help later this week or next one

PR #1138, opened 2025-12-05
Feature consul tls discovery

chc0618:
### Background

We are using SmallRye Stork with Consul in production.
In our production environment, Consul is running with TLS enabled.

Currently, Stork Consul discovery does not support TLS configuration,
so we added SSL/TLS support and verified it in production.

### What’s included

- Support TLS when discovery services to Consul
- Support configuration for:
  - `trust-store-path` (CA file)
  - `trust-store-password`
  - `key-store-path` (client certificate + private key)
  - `key-store-password`
  - `verify-host` (hostname verification)
  - `acl-token` (Consul ACL authentication)
- Backward compatible (non-TLS users are not affected)

### Production status

✅ Already running in production.

### Why this is useful

More and more users deploy Consul in secure (TLS) mode.
This change allows Stork to be used in secure enterprise environments.


service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryProvider.java
```diff
@@ -20,6 +20,13 @@
 @ServiceDiscoveryAttribute(name = "application", description = "The application name; if not defined Stork service name will be used.")
 @ServiceDiscoveryAttribute(name = "refresh-period", description = "Service discovery cache refresh period.", defaultValue = CachingServiceDiscovery.DEFAULT_REFRESH_INTERVAL)
 @ServiceDiscoveryAttribute(name = "secure", description = "whether the connection with the service should be encrypted with TLS.")
+@ServiceDiscoveryAttribute(name = "ssl", description = "是否使用 ssh", defaultValue = "false")
+@ServiceDiscoveryAttribute(name = "trust-store-path", description = "trust-store-path", defaultValue = "")
+@ServiceDiscoveryAttribute(name = "trust-store-password", description = "trust-store-password", defaultValue = "")
+@ServiceDiscoveryAttribute(name = "key-store-path", description = "key-store-path", defaultValue = "")
+@ServiceDiscoveryAttribute(name = "key-store-password", description = "key-store-password", defaultValue = "")
+@ServiceDiscoveryAttribute(name = "verify-host", description = "verify-host", defaultValue = "false")
```
aureamunoz: I'm wondering if it would be better to have `verify-host=true` by default, wdyt?
cc @cescoffier

chc0618: By default, hostname verification is disabled to maintain compatibility with common internal deployments (IP-based access, self-signed certificates, Kubernetes environments). Users can explicitly enable it for stricter TLS security when using proper DNS and CA-issued certificates.

service-discovery/consul/src/main/java/io/smallrye/stork/servicediscovery/consul/ConsulServiceDiscoveryProvider.java
```diff
@@ -20,6 +20,13 @@
 @ServiceDiscoveryAttribute(name = "application", description = "The application name; if not defined Stork service name will be used.")
 @ServiceDiscoveryAttribute(name = "refresh-period", description = "Service discovery cache refresh period.", defaultValue = CachingServiceDiscovery.DEFAULT_REFRESH_INTERVAL)
 @ServiceDiscoveryAttribute(name = "secure", description = "whether the connection with the service should be encrypted with TLS.")
+@ServiceDiscoveryAttribute(name = "ssl", description = "是否使用 ssh", defaultValue = "false")
```
aureamunoz: Also we need to put description in English

PR #1146, opened 2025-12-19
Support k8s endpoint slice api

aureamunoz:


cescoffier (dismissed):
I would mark it as experimental for now.

cescoffier (approved):

PR #1204, opened 2026-04-21
Filter non stork configs and read them as optional values

### LINKED ISSUE #1201: Microprofile Provider: Stork should only call config.getValue() on properties related to Stork
Initializing Stork results in an exception if there are any configuration properties with empty strings as values.

**Code to reproduce**
```java
SmallRyeConfig config = new SmallRyeConfigBuilder()
      .addDefaultSources()
      .build();

ConfigProviderResolver.instance().registerConfig(config, null);

Stork.initialize();
```
If there are empty system properties or environmental variables, the stork `MicroProfileConfigProvider` will throw an exception as it eagerly reads the configuration values for all existing properties:

[**Source**](https://github.com/smallrye/smallrye-stork/blob/main/microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java)
```java
for (String propertyName : config.getPropertyNames()) {
    StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
            config.getValue(propertyName, String.class));
}
```

**The resulting exception**

`SRCFG00040: The config property user.script is defined as the empty String ("") which the following Converter considered to be null: io.smallrye.config.Converters$BuiltInConverter`

Currently I have solved this issue by adding a custom interceptor to the config provider, but I don't see why Stork needs to read the values of all properties, it really should filter them and only read the ones related to Stork (`stork.*`).

**Versions I have used:**

- JDK 21 Eclipse Temurin 21.0.4
- Maven 3.9.12
- io.smallrye.config:smallrye-config-core 3.16.0
- io.smallrye.stork:stork-core 2.7.9
- io.smallrye.stork:stork-microprofile-config 2.7.9

aureamunoz:
Fixes #1201

### DIFF
--- a/microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java
+++ b/microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java
@@ -9,6 +9,7 @@ import org.eclipse.microprofile.config.Config;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.smallrye.stork.Stork;
 import io.smallrye.stork.api.config.ServiceConfig;
 import io.smallrye.stork.spi.config.ConfigProvider;
 import io.smallrye.stork.spi.config.SimpleServiceConfig;
@@ -32,8 +33,15 @@ public class MicroProfileConfigProvider implements ConfigProvider {
         Map<String, Map<String, String>> propertiesByServiceName = new HashMap<>();
 
         for (String propertyName : config.getPropertyNames()) {
-            StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
-                    config.getValue(propertyName, String.class));
+            if (propertyName.startsWith(Stork.STORK + ".")) {
+                config.getOptionalValue(propertyName, String.class)
+                        .map(String::trim)
+                        .filter(value -> !value.isEmpty())
+                        .ifPresentOrElse(
+                                value -> StorkConfigUtils.computeServiceProperty(propertiesByServiceName,
+                                        propertyName, value),
+                                () -> log.debug("Ignoring empty stork config property: {}", propertyName));
+            }
         }
 
         for (Map.Entry<String, Map<String, String>> serviceEntry : propertiesByServiceName.entrySet()) {
--- a/microprofile/src/test/java/io/smallrye/stork/microprofile/MicroProfileConfigProviderTest.java
+++ b/microprofile/src/test/java/io/smallrye/stork/microprofile/MicroProfileConfigProviderTest.java
@@ -233,6 +233,28 @@ public class MicroProfileConfigProviderTest {
         assertThat(lb.getType()).isEqualTo("test-lb-2");
     }
 
+    @Test
+    void shouldFilterNonStorkPropertiesAndIgnoreEmptyValues() {
+        Map<String, String> properties = new HashMap<>();
+        properties.put("stork." + FIRST_SERVICE + ".service-discovery", "test-sd-1");
+        properties.put("stork." + FIRST_SERVICE + ".service-discovery.one", "http://localhost:8080");
+        properties.put("stork." + FIRST_SERVICE + ".service-discovery.two", "");
+        properties.put("stork." + FIRST_SERVICE + ".service-discovery.three", "   ");
+        properties.put("user.script", "");
+        properties.put("some.other.property", "");
+        properties.put("non.stork.property", "some-value");
+
+        Stork stork = storkForConfig(properties);
+
+        ServiceDiscovery serviceDiscovery = stork.getService(FIRST_SERVICE).getServiceDiscovery();
+        assertThat(serviceDiscovery).isNotNull().isInstanceOf(TestServiceDiscovery.class);
+
+        TestServiceDiscovery sd = (TestServiceDiscovery) serviceDiscovery;
+        assertThat(sd.getType()).isEqualTo("test-sd-1");
+        assertThat(sd.getConfig().getOne()).isEqualTo("http://localhost:8080");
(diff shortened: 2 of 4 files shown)


microprofile/src/test/java/io/smallrye/stork/microprofile/MicroProfileConfigProviderTest.java
```diff
@@ -233,6 +233,27 @@ void shouldHandleServiceNamesInQuotes() {
         assertThat(lb.getType()).isEqualTo("test-lb-2");
     }
 
+    @Test
+    void shouldIgnorePropertiesWithEmptyValues() {
```
cescoffier: Something like `shouldFilterNonStorkPropertiesAndIgnoreEmptyValues` would be sightly better.

microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java
```diff
@@ -32,8 +33,11 @@ public MicroProfileConfigProvider() {
         Map<String, Map<String, String>> propertiesByServiceName = new HashMap<>();
 
         for (String propertyName : config.getPropertyNames()) {
-            StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
-                    config.getValue(propertyName, String.class));
+            if (propertyName.startsWith(Stork.STORK + ".")) {
+                config.getOptionalValue(propertyName, String.class)
+                        .ifPresent(value -> StorkConfigUtils.computeServiceProperty(propertiesByServiceName,
```
cescoffier: Maybe add ` log.debug("Ignoring empty stork config property: {}", propertyName)` just in case.

cescoffier (dismissed):
Just 2 minor comments.

cescoffier:
The Spring Boot config provider doesn't have the equivalent filtering / optional value handling:

```java
for (String propertyName : getPropertyNames(environment)) {
    StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
            environment.getProperty(propertyName));
}
```

Should we add it there too?

cescoffier (dismissed):

aureamunoz:
> The Spring Boot config provider doesn't have the equivalent filtering / optional value handling:
> 
> ```java
> for (String propertyName : getPropertyNames(environment)) {
>     StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
>             environment.getProperty(propertyName));
> }
> ```
> 
> Should we add it there too?

Yes, I think so for consistency. I saw it at some point but didn't do. It's done in last commit.

spring-boot/src/main/java/io/smallrye/stork/springboot/SpringBootConfigProvider.java
```diff
@@ -26,10 +27,12 @@ public SpringBootConfigProvider() {
...
 
         for (String propertyName : getPropertyNames(environment)) {
-
-            StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
-                    environment.getProperty(propertyName));
-
+            if (propertyName.startsWith(Stork.STORK + ".")) {
+                String value = environment.getProperty(propertyName);
+                if (value != null && !value.isEmpty()) {
+                    StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName, value);
```
cescoffier: Should we log otherwise, as we do in MP?

spring-boot/src/main/java/io/smallrye/stork/springboot/SpringBootConfigProvider.java
```diff
@@ -26,10 +27,12 @@ public SpringBootConfigProvider() {
         Map<String, Map<String, String>> propertiesByServiceName = new HashMap<>();
 
         for (String propertyName : getPropertyNames(environment)) {
-
-            StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
-                    environment.getProperty(propertyName));
-
+            if (propertyName.startsWith(Stork.STORK + ".")) {
+                String value = environment.getProperty(propertyName);
+                if (value != null && !value.isEmpty()) {
```
cescoffier: Wondering if we should trim the value first and avoid 0-length values. Not sure.

microprofile/src/main/java/io/smallrye/stork/microprofile/MicroProfileConfigProvider.java
```diff
@@ -32,8 +33,13 @@ public MicroProfileConfigProvider() {
...
 
         for (String propertyName : config.getPropertyNames()) {
-            StorkConfigUtils.computeServiceProperty(propertiesByServiceName, propertyName,
-                    config.getValue(propertyName, String.class));
+            if (propertyName.startsWith(Stork.STORK + ".")) {
+                config.getOptionalValue(propertyName, String.class)
+                        .ifPresentOrElse(
+                                value -> StorkConfigUtils.computeServiceProperty(propertiesByServiceName,
+                                        propertyName, value),
+                                () -> log.debug("Ignoring empty Stork config property: {}", propertyName));
```
cescoffier: Stork or stork? Our logging seems to be slightly inconsistent.

cescoffier (dismissed):
See a few minor comments, but nothing is holding the merge.

spring-boot/src/main/java/io/smallrye/stork/springboot/SpringBootConfigProvider.java
aureamunoz: Yes, let's do it I think it's a good idea. It doesn't cost anything, and a value with blank spaces isn't usually intentional.

aureamunoz:
I need your approval again to merge

cescoffier (approved):

Issue #1209, opened 2026-04-21
Kubernetes service discovery with server-side load balancing?


aureamunoz:
### Discussed in https://github.com/smallrye/smallrye-stork/discussions/1035

<div type='discussions-op-text'>

<sup>Originally posted by **mseitner** April 23, 2025</sup>
Hello,

If I understand Stork's built-in support for Kubernetes services correctly then it tracks the endpoints of a service and performs client-side load balancing.

Is it possible to instruct Stork to use a service's cluster IP and skip client-side LB?

Some background: I have a multi-port K8s service and want Stork to only do the name-based service and port discovery. This way I could easily configure where the load balancing should happen without having to alter my application code.</div>

aureamunoz:
Implemented by #1210

PR #1210, opened 2026-04-22
Add use-cluster-ip option to Kubernetes service discovery

aureamunoz:
Resolves to the Service ClusterIP and port instead of individual pod IPs, enabling service-level resolution without client-side load balancing.

Related to #1209


service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -300,6 +331,69 @@ private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
         return slicesUni;
     }
 
+    private Uni<List<ServiceInstance>> fetchServiceClusterIp(List<ServiceInstance> previousInstances) {
+        return Uni.createFrom().emitter(
+                emitter -> {
+                    vertx.executeBlocking(future -> {
```
cescoffier: What is we have allNamespaces + useClusterIp?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -300,6 +331,69 @@ private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
...
+                        }
+                        String clusterIp = svc.getSpec().getClusterIP();
+                        if (clusterIp == null || "None".equals(clusterIp)) {
+                            future.fail(new IllegalStateException(
+                                    "Service '" + application + "' has no ClusterIP (headless service)"));
+                            return;
+                        }
+
+                        List<ServicePort> servicePorts = svc.getSpec().getPorts();
+                        int port = 0;
```
cescoffier: First use -1, then, we need to check we get a port.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -300,6 +331,69 @@ private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
...
+                            return;
+                        }
+                        String clusterIp = svc.getSpec().getClusterIP();
+                        if (clusterIp == null || "None".equals(clusterIp)) {
+                            future.fail(new IllegalStateException(
+                                    "Service '" + application + "' has no ClusterIP (headless service)"));
+                            return;
+                        }
+
+                        List<ServicePort> servicePorts = svc.getSpec().getPorts();
```
cescoffier: If a Service has no ports defined (unusual but possible), this will be null and fails with an NPE later.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -300,6 +331,69 @@ private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
         return slicesUni;
     }
 
+    private Uni<List<ServiceInstance>> fetchServiceClusterIp(List<ServiceInstance> previousInstances) {
```
cescoffier: There is some duplicated logic between this method and fromEndpointsToStorkServiceInstances (port selection).

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -300,6 +331,69 @@ private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
...
+                        String clusterIp = svc.getSpec().getClusterIP();
+                        if (clusterIp == null || "None".equals(clusterIp)) {
+                            future.fail(new IllegalStateException(
+                                    "Service '" + application + "' has no ClusterIP (headless service)"));
+                            return;
+                        }
+
+                        List<ServicePort> servicePorts = svc.getSpec().getPorts();
+                        int port = 0;
+                        String protocol = "";
```
cescoffier: Use `null`, and make sure we get a value.

cescoffier (changes requested):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Check + log added

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: refactored

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: If, following the port resolution, port == -1 (no port matches portName), a warning is logged and the process continues. This prevents an instance from being created with port -1 and no protocol

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
```diff
@@ -24,6 +24,7 @@
 @ServiceDiscoveryAttribute(name = "request-retry-backoff-limit", description = "Maximum number of retry attempts allowed after a request failure.")
 @ServiceDiscoveryAttribute(name = "request-retry-backoff-interval", description = "Amount of time to wait between retry attempts after a request fails.")
 @ServiceDiscoveryAttribute(name = "use-endpoint-slices", description = "Use the EndpointSlice API instead", defaultValue = "false")
+@ServiceDiscoveryAttribute(name = "use-cluster-ip", description = "Use the Kubernetes Service ClusterIP instead of pod IPs. Stork returns a single instance with the ClusterIP and resolved port.", defaultValue = "false")
```
cescoffier, line 27: Did you add a test for allNamespace + useClusterIP ?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -382,55 +446,69 @@ private List<ServiceInstance> fromEndpointsToStorkServiceInstances(Map<Endpoints
...
                     }
                     List<EndpointPort> endpointPorts = subset.getPorts();
-                    Integer port = 0;
-                    String protocol = "";
+                    if (endpointPorts == null || endpointPorts.isEmpty()) {
+                        continue;
+                    }
+                    Integer port = -1;
+                    String protocol = null;
                     if (endpointPorts.size() == 1) {
```
cescoffier: This snippet is still duplicated several times. It's not long or overly complex, but it would be better to isolate and reuse it.

cescoffier: ah, in one case it's a ServicePort and the other one EndpointPort. Do they share a common interface?

docs/docs/service-discovery/kubernetes.md
```diff
@@ -162,6 +162,85 @@ If different port combinations exist, Kubernetes exposes them as separate Endpoi
...
+In this mode, Stork queries the Kubernetes `Service` resource directly and returns a single `ServiceInstance` with:
+
+* **Host** — the Service's ClusterIP (e.g., `10.96.0.15`)
+* **Port** — resolved from the Service spec, matching `port-name` when multiple ports are defined
+
+Since only one instance is returned, any configured Stork load balancer becomes a passthrough.
+
+!!! note
+    ClusterIP mode does not support headless Services (`clusterIP: None`). 
+If the target Service is headless, Stork will fail with an error and fall back to the last known instances (or an empty list on first call).
```
cescoffier: Is it true that the code does not fail, just warns and skips?

cescoffier (changes requested):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
aureamunoz: done

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Refactored using generics and same pattern that for the informer

docs/docs/service-discovery/kubernetes.md
aureamunoz: Yes, and the docs says the opposite. I've fixed it so that when no service instances are resolved (across all three modes: ClusterIP, Endpoints, and EndpointSlices), Stork now fails with an error. This triggers the CachingServiceDiscovery fallback, which returns the last known instances instead of an empty list. The corresponding test have been added/adjusted accordingly.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -108,7 +112,14 @@ public KubernetesServiceDiscovery(String serviceName, KubernetesConfiguration co
         this.secure = isSecure(config);
         this.useEndpointSlices = config.getUseEndpointSlices() == null ? null
                 : Boolean.valueOf(config.getUseEndpointSlices());
-        if (shouldUseEndpointSlices()) {
+        this.useClusterIp = config.getUseClusterIp() != null && Boolean.parseBoolean(config.getUseClusterIp());
+        if (useClusterIp && useEndpointSlices) {
```
cescoffier: There is an NPE here.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -218,87 +247,124 @@ public void invalidate() {
...
         } else {
-            Uni<Map<Endpoints, List<Pod>>> endpointsUni = fetchServiceEnpoints();
-            return endpointsUni.onItem()
-                    .transform(endpoints -> fromEndpointsToStorkServiceInstances(endpoints, previousInstances))
-                    .invoke(() -> invalidated.set(false));
+            result = fetchServiceEnpoints().onItem()
+                    .transform(endpoints -> fromEndpointsToStorkServiceInstances(endpoints, previousInstances));
         }
+        return result
+                .onItem().transformToUni(this::failIfEmpty)
```
cescoffier: Hum, this is changing all "modes".

Previously, when all endpoints were deleted, Stork returned an empty list. Now it fails, triggering `CachingServiceDiscovery` fallback to last-known instances. Is it what you want?

cescoffier (changes requested):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Yes, I know it is changing all modes, I did it for consistency. Also it is what documentation describes. That makes sense to me in case of temporary errors for example, the previous behavior was failing silently and returning an empty list in that case.

aureamunoz: I finally reverted this behavior in favor of the old one. It seems better to return an empty list, we already have a cache mechanism in place. The issue https://github.com/smallrye/smallrye-stork/issues/1232 has been created for enable this feature by configuration.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -218,87 +247,114 @@ public void invalidate() {
...
 
     private Uni<List<EndpointSlice>> fetchServiceEndpointSlice() {
-        Uni<List<EndpointSlice>> slicesUni = Uni.createFrom().emitter(
-                emitter -> {
-                    vertx.executeBlocking(future -> {
-                        List<EndpointSlice> items = new ArrayList<>();
+        return executeOnWorkerThread(() -> {
+            if (allNamespaces) {
+                return client.discovery().v1().endpointSlices()
+                        .inNamespace(namespace)
```
cescoffier, line 302: I'm confused - both branches are identical, and both pass "namespace".

cescoffier: shouldn't this be `inAnyNamespace()`?  (we need tests to make sure it works).

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -383,55 +432,73 @@ private List<ServiceInstance> fromEndpointsToStorkServiceInstances(Map<Endpoints
...
         return serviceInstances;
     }
 
+    private record ResolvedPort(int port, String protocol) {
+    }
+
+    private <P> ResolvedPort resolvePort(List<P> ports,
+            Function<P, String> getName, Function<P, Integer> getPort, Function<P, String> getProtocol) {
+        if (ports.size() == 1) {
+            return new ResolvedPort(getPort.apply(ports.get(0)), getProtocol.apply(ports.get(0)));
```
cescoffier, line 479: Hum, and what if it does not match?

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -218,87 +247,114 @@ public void invalidate() {
...
-                            emitter.complete(endpointSlices);
-                        } else {
-                            LOGGER.error("Unable to retrieve the endpoint from the {} service", application,
-                                    result.cause());
-                            emitter.fail(result.cause());
-                        }
-                    });
-                });
-        return slicesUni;
+    List<ServiceInstance> fromClusterIpServicesToStorkServiceInstances(List<Service> clusterIpServices,
```
cescoffier: why package private?

docs/docs/service-discovery/kubernetes.md
```diff
@@ -161,6 +161,85 @@ If different port combinations exist, Kubernetes exposes them as separate Endpoi
...
+
+In this mode, Stork queries the Kubernetes `Service` resource directly and returns a single `ServiceInstance` with:
+
+* **Host** — the Service's ClusterIP (e.g., `10.96.0.15`)
+* **Port** — resolved from the Service spec, matching `port-name` when multiple ports are defined
+
+Since only one instance is returned, any configured Stork load balancer becomes a passthrough.
+
+!!! note
+    ClusterIP mode does not support headless Services (`clusterIP: None`).
```
cescoffier, line 206: Why the indentation?

cescoffier: Actually, I think it needs indentation on the whole block.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -218,87 +247,114 @@ public void invalidate() {
...
+                    .transform(clusterIps -> fromClusterIpServicesToStorkServiceInstances(clusterIps, previousInstances));
+        } else if (shouldUseEndpointSlices()) {
+            result = fetchServiceEndpointSlice().onItem()
+                    .transform(slices -> fromSlicesToStorkServiceInstances(slices, previousInstances));
         } else {
-            Uni<Map<Endpoints, List<Pod>>> endpointsUni = fetchServiceEnpoints();
-            return endpointsUni.onItem()
-                    .transform(endpoints -> fromEndpointsToStorkServiceInstances(endpoints, previousInstances))
-                    .invoke(() -> invalidated.set(false));
+            result = fetchServiceEnpoints().onItem()
```
cescoffier:
```suggestion
            result = fetchServiceEndpoints().onItem()
```

cescoffier (changes requested):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Oops! Good catch! That’s a bug: this method returns *EndpointSlice* objects, not *ClusterIP* services (see line 316 – I was getting crazy because I remembered implementing it for *ClusterIP*), which is the aim of this pull request. I’ll fix it in a new branch or a specific pull request.

aureamunoz: https://github.com/smallrye/smallrye-stork/issues/1235 created for fixing this

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: It returns null. And the methods that call it already have a safeguard in place for this: they execute a ‘continue’ if `resolvePort` returns null, skipping that instance without adding it to the list of results.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: I didn't pay attention. I will add `private`key word

aureamunoz:
Can we merge this one?

cescoffier (approved):

Issue #1211, opened 2026-04-27
Kubernetes service-discovery use-endpoint-slices not working as expected


rikcarve:
Hi all

The support for Kubernetes EndpointSlices always queries the kubernetes API for endpointslices, even when explicitly disabled in the config. This is an issue as in some environments, the service account may not have the rights to do so leading to an error.

From what I see in the code, the shouldUseEndpointSlices() method in KubernetesServiceDiscovery.java does this query even though it already knows it's disabled in the config. The method should immediately return false in case it was disabled in the config.
https://github.com/smallrye/smallrye-stork/blob/6d1c6602dd868b4faa57fb8cefd10bf7d088fa8f/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java#L136

```
 io.quarkus.runner.GeneratedMain.main(Unknown Source)
	at io.quarkus.bootstrap.runner.QuarkusEntryPoint.doRun(QuarkusEntryPoint.java:86)
	at io.quarkus.bootstrap.runner.QuarkusEntryPoint.main(QuarkusEntryPoint.java:37)
	  Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://172.31.0.1:443/apis/discovery.k8s.io/v1/namespaces/replaced/endpointslices?labelSelector=kube…
	  Message: endpointslices.discovery.k8s.io is forbidden: User \"system:serviceaccount:replaced:default\" cannot list resource \"endpointslices\" in API group \"discovery.k8s.io\" in the namespace \"replaced\". Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=discovery.k8s.io, kind=endpointslices, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=endpointslices.discovery.k8s.io is forbidden: User \"system:serviceaccount:replaced:default\" cannot list resource \"endpointslices\" in API group \"discovery.k8s.io\" in the namespace \"replaced\", metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, status=Failure, additionalProperties={}).
	at io.fabric8.kubernetes.client.KubernetesClientException.copyAsCause(KubernetesClientException.java:205)
[...]

PR #1220, opened 2026-04-30
fix: do not autodetect use-endpoint-slices if disabled in config

rikcarve:
Do not use autodetect for use-endpoint-slices when explicitly disabled in the config.

Autodetect is now done only when config is not set (default) or config is set to true (to avoid later errors). 
`use-endpoint-slices` has now no default value (was false before)

#1211


service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
```diff
@@ -23,7 +23,7 @@
 @ServiceDiscoveryAttribute(name = "secure", description = "Whether the connection with the service should be encrypted with TLS.")
 @ServiceDiscoveryAttribute(name = "request-retry-backoff-limit", description = "Maximum number of retry attempts allowed after a request failure.")
 @ServiceDiscoveryAttribute(name = "request-retry-backoff-interval", description = "Amount of time to wait between retry attempts after a request fails.")
-@ServiceDiscoveryAttribute(name = "use-endpoint-slices", description = "Use the EndpointSlice API instead", defaultValue = "false")
+@ServiceDiscoveryAttribute(name = "use-endpoint-slices", description = "Use the EndpointSlice API instead")
```
cescoffier, line 26: Don't we have a deprecated attribute?

cescoffier (changes requested):
I would like to see a test for the new behavior - the rest looks great! Thanks!

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
rikcarve: Not sure what you mean with "deprecated attribute".

As for the tests, all the tests were already there, and should have failed with the old code. I adapted them slightly so that they would fail with the previous code, but are now green. Please check them, as I'm not 100% sure...

aureamunoz (commented):
Thanks for the fix! Just a minor comment: Could you pleas also add to the documentation that when explicitly set, this configuration always takes priority over automatic detection?

I have taken the liberty of modifying the `shouldUseEndpointsWhenUserDisablesEndpointSlices` test (and push it to your branch), which was passing despite the error. This test needs to register EndpointSlices (with a different IP address) so that we can verify that, even though EndpointSlices are available, the code respects the `use-endpoint-slices=false` configuration and uses the classic Endpoints.

aureamunoz (dismissed):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryProvider.java
rikcarve: I think the current documentation is a bit misleading and did not match the implementation. Let me try to adapt it

rikcarve: Guess I need some help with the API thing failing...

aureamunoz: Let me check

aureamunoz: I merged all in one, I think it wasn't happy because of 2 entries for the same attribute but not sure what was the problem. Let's see what CI says.

aureamunoz (approved):

aureamunoz:
CI is green. Could yo please squash commits? Then it should be ok to merge :-)

rikcarve:
Squash done

aureamunoz (approved):

PR #1223, opened 2026-05-04
Add permissions to post-release workflow

gastaldi:
The workflow updates GitHub release notes, which requires
contents: write. Also switch from RELEASE_TOKEN to the built-in
GITHUB_TOKEN now that the correct permission scope is declared.

Use jbangdev/setup-jbang action instead of manual sdkman installation.


cescoffier (approved):

cescoffier:
Thanks!

PR #1233, opened 2026-05-06
Fix filter ports by name when using slices in k8s sd

### LINKED ISSUE #1230: EndpointSlice service discovery ignores port-name filter
## Description

The `fromSlicesToStorkServiceInstances()` method introduced in PR #1146 does not respect the `port-name` configuration. It creates `ServiceInstance` objects for **all ports** in an EndpointSlice, not just the port matching the configured `port-name`.

This is a regression from the Endpoints API path (`fromEndpointsToStorkServiceInstances()`), which correctly filters ports by name.

## Impact

When a Kubernetes Service exposes multiple ports (e.g., `http` for gRPC on port 8080 and `management` for metrics on port 9090), Stork creates ServiceInstance entries for **both ports**. The gRPC/HTTP client then randomly load-balances between them, causing ~50% of requests to fail when they hit the wrong port.

## Root Cause

**File:** `service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java`  
**Method:** `fromSlicesToStorkServiceInstances(List<EndpointSlice>, List<ServiceInstance>)`  
**Lines:** ~300-330 (main branch)

### Current buggy code:

```java
for (String address : endpoint.getAddresses()) {
    for (io.fabric8.kubernetes.api.model.discovery.v1.EndpointPort port : ports) {
        // ❌ NO port-name filter - creates instance for EVERY port
        ServiceInstance serviceInstance = new DefaultServiceInstance(
            ServiceInstanceIds.next(),
            address, 
            port.getPort(),  // Uses ALL ports
            Optional.empty(), 
            secure,
            labels,
            k8sMetadata
        );
        serviceInstances.add(serviceInstance);
    }
}
```

### Expected behavior (from working Endpoints path):

```java
for (EndpointPort port : ports) {
    // ✅ Filters by port-name
    if (portName != null && portName.equals(port.getName())) {
        selectedPort = port.getPort();
        break;  // Stop after finding match
    }
}
// Create instance with ONLY the matching port
```

## Reproduction

1. Create a Kubernetes Service with multiple named ports:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  ports:
    - name: http
      port: 8080
      targetPort: 8080
    - name: management
      port: 9090
      targetPort: 9090
```

2. Configure Stork with `port-name=http`:
```properties
quarkus.stork.my-service.service-discovery.type=kubernetes
quarkus.stork.my-service.service-discovery.port-name=http
quarkus.stork.my-service.service-discovery.k8s-namespace=default
quarkus.stork.my-service.service-discovery.application=my-service
```

3. Enable EndpointSlice support (or let it auto-detect):
```properties
quarkus.stork.my-service.service-discovery.use-endpoint-slices=true
```

4. Observe that Stork discovers instances on **both ports 8080 and 9090** instead of just 8080.

## Expected Behavior

Only the port matching `port-name=http` should be used for service discovery.

## Actual Behavior

All ports are used, ignoring the `port-name` filter.

## Environment

- **Stork version:** 2.7.9 (latest)
- **Quarkus version:** 3.33.1.1
- **Kubernetes version:** 1.29
- **Affected code:** `main` branch (confirmed as of 2026-05-05)

## Workaround

Disable EndpointSlice support to force the working Endpoints API path:

```properties
quarkus.stork.my-service.service-discovery.use-endpoint-slices=false
```

## Suggested Fix

Add port-name filtering in `fromSlicesToStorkServiceInstances()`:

```java
for (String address : endpoint.getAddresses()) {
    for (io.fabric8.kubernetes.api.model.discovery.v1.EndpointPort port : ports) {
        // Filter by port-name if configured
        if (portName != null && !portName.equals(port.getName())) {
            continue;  // Skip ports that don't match
        }
        
        ServiceInstance matching = ServiceInstanceUtils.findMatching(previousInstances, address, port.getPort());
        if (matching != null) {
            serviceInstances.add(matching);
        } else {
            // ... create new instance
        }
    }
}
```

This matches the behavior in `fromEndpointsToStorkServiceInstances()` which already implements this correctly.

## Related

- PR #1146 - Introduced EndpointSlice support (where this bug was introduced)
- PR #547 - Added `port-name` feature (works correctly for Endpoints API)
- Issue #1211 - Different EndpointSlice bug (auto-detection when disabled)

aureamunoz:
Fixes #1230 

cc [Sirz3chs](https://github.com/Sirz3chs)

### DIFF
--- a/service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryWithSlicesTest.java
+++ b/service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryWithSlicesTest.java
@@ -398,6 +398,38 @@ public class KubernetesServiceDiscoveryWithSlicesTest {
         });
     }
 
+    @Test
+    void shouldReturnNoInstancesWhenPortNameMatchesNothing() {
+        String serviceName = "svc";
+
+        TestConfigProvider.addServiceConfig(serviceName, null, "kubernetes", null,
+                null, Map.of("k8s-host", k8sMasterUrl,
+                        "k8s-namespace", "ns1",
+                        "port-name", "metrics",
+                        "use-endpoint-slices", "true"),
+                null);
+        Stork stork = StorkTestUtils.getNewStorkInstance();
+
+        whenGetSlicesApiAvailableThenReturnTrue();
+
+        String[] ips = { "10.96.96.231", "10.96.96.232" };
+        EndpointPort[] ports = { new EndpointPort("http", 8080) };
+
+        registerKubernetesEndpointSlice(serviceName, "ns1", ips, ports);
+
+        AtomicReference<List<ServiceInstance>> instances = new AtomicReference<>();
+
+        Service service = stork.getService(serviceName);
+        service.getServiceDiscovery().getServiceInstances()
+                .onFailure().invoke(th -> fail("Failed to get service instances from Kubernetes", th))
+                .subscribe().with(instances::set);
+
+        await().atMost(Duration.ofSeconds(5))
+                .until(() -> instances.get() != null);
+
+        assertThat(instances.get()).hasSize(0);
+    }
+
     private void whenGetSlicesApiAvailableThenReturnTrue() {
         server.expect()
                 .get()
@@ -495,6 +527,6 @@ public class KubernetesServiceDiscoveryWithSlicesTest {
     }
 
     record EndpointPort(String name, int portNumber) {
-    };
+    }
 
 }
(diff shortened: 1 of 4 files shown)


service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryWithSlicesTest.java
```diff
@@ -448,4 +494,7 @@ private void registerKubernetesEndpointSlice(String serviceName, String namespac
 
     }
 
+    record EndpointPort(String name, int portNumber) {
+    };
```
cescoffier: Cosmetic: I don't believe the semi-colon is necessary.

cescoffier (dismissed):
Before merging, could you add a test verifying the behavior when the port-name is set to a non existent port name? For example:
```
 @Test
  void shouldReturnNoInstancesWhenPortNameMatchesNothing() {
      // Configure port-name="grpc" but only have "http" and "management" ports
      // Assert: instances list is empty
  }
```

PR #1242, opened 2026-05-13
 Fix multiple instances overriding each other when registering with consul

### LINKED ISSUE #1241: Consul service registration overwrites instances when multiple instances share the same service name
When registering multiple instances of the same service, Consul uses the service name as the consulId. Since all instances share the same service name, each new registration overwrites the previous one instead of registering as a separate instance.

aureamunoz:
Add instanceName support to ServiceRegistrar

Introduce registerServiceInstance(serviceName, instanceName, metadata, ip, port) as a new abstract method on ServiceRegistrar so callers can supply an explicit instance ID instead of the auto-generated serviceName::ip::port format. The existing 4-arg signature becomes a default that delegates to the new 5-arg with null, preserving backwards compatibility.
Also adds deregisterServiceInstance(serviceName, instanceName) to allow deregistration by an explicit ID, with concrete implementations in Consul and Eureka.

Fixes #1241

### DIFF
--- a/api/src/main/java/io/smallrye/stork/api/Metadata.java
+++ b/api/src/main/java/io/smallrye/stork/api/Metadata.java
@@ -2,6 +2,7 @@ package io.smallrye.stork.api;
 
 import java.util.Collections;
 import java.util.EnumMap;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -95,6 +96,18 @@ public class Metadata<T extends Enum<T> & MetadataKey> {
      */
     public EnumMap<T, Object> getMetadata() {
         return metadata;
+    }
+
+    /**
+     * @return the metadata as a {@code Map<String, String>}, using each key's name and converting values via
+     *         {@code toString()}
+     */
+    public Map<String, String> asMap() {
+        Map<String, String> result = new HashMap<>();
+        for (Map.Entry<T, Object> entry : metadata.entrySet()) {
+            result.put(entry.getKey().getName(), entry.getValue() != null ? entry.getValue().toString() : null);
+        }
+        return result;
     }
 
     /**
--- a/api/src/main/java/io/smallrye/stork/api/Service.java
+++ b/api/src/main/java/io/smallrye/stork/api/Service.java
@@ -231,16 +231,75 @@ public class Service {
         return serviceRegistrar;
     }
 
+    /**
+     * Registers a service instance identified by its IP address and port.
+     *
+     * @param serviceName the name of the service
+     * @param ipAddress the IP address of the instance
+     * @param port the port of the instance
+     * @return a {@link Uni} completing when the registration is done
+     */
     public Uni<Void> registerInstance(String serviceName, String ipAddress, int port) {
         return serviceRegistrar.registerServiceInstance(serviceName, ipAddress, port);
     }
 
+    /**
+     * Registers a service instance with an explicit instance name used as the registration ID.
+     *
+     * @param serviceName the name of the service
+     * @param instanceName the unique identifier for this instance in the registry
+     * @param ipAddress the IP address of the instance
+     * @param port the port of the instance
+     * @return a {@link Uni} completing when the registration is done
+     */
+    public Uni<Void> registerInstance(String serviceName, String instanceName, String ipAddress, int port) {
+        return serviceRegistrar.registerServiceInstance(serviceName, instanceName, Metadata.empty(), ipAddress, port);
+    }
+
+    /**
+     * Registers a service instance using the provided options (tags, metadata, etc.).
+     *
+     * @param options the registration options
+     * @return a {@link Uni} completing when the registration is done
+     */
     public Uni<Void> registerInstance(ServiceRegistrar.RegistrarOptions options) {
         return serviceRegistrar.registerServiceInstance(options);
     }
 
+    /**
+     * Deregisters all instances of the given service.
+     *
+     * @param serviceName the name of the service
+     * @return a {@link Uni} completing when the deregistration is done
+     * @deprecated Prefer {@link #deregisterServiceInstance(String, String)} or
+     *             {@link #deregisterServiceInstance(String, String, int)} to target a specific instance.
+     */
+    @Deprecated
     public Uni<Void> deregisterServiceInstance(String serviceName) {
         return serviceRegistrar.deregisterServiceInstance(serviceName);
+    }
+
+    /**
+     * Deregisters the service instance identified by the given IP address and port.
+     *
+     * @param serviceName the name of the service
+     * @param ipAddress the IP address of the instance to deregister
+     * @param port the port of the instance to deregister
+     * @return a {@link Uni} completing when the deregistration is done
+     */
+    public Uni<Void> deregisterServiceInstance(String serviceName, String ipAddress, int port) {
+        return serviceRegistrar.deregisterServiceInstance(serviceName, ipAddress, port);
+    }
+
+    /**
+     * Deregisters the service instance identified by the given instance name.
+     *
+     * @param serviceName the name of the service
+     * @param instanceName the registration ID of the instance to deregister
+     * @return a {@link Uni} completing when the deregistration is done
+     */
+    public Uni<Void> deregisterServiceInstance(String serviceName, String instanceName) {
+        return serviceRegistrar.deregisterServiceInstance(serviceName, instanceName);
     }
 
     public ObservationCollector getObservations() {
--- a/api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
+++ b/api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
@@ -18,14 +18,25 @@ public interface ServiceRegistrar<MetadataKeyType extends Enum<MetadataKeyType> 
         }
     }
 
+    default void checkInstanceNameNotNull(String instanceName) {
+        if (instanceName == null || instanceName.isEmpty() || instanceName.isBlank()) {
+            throw new IllegalArgumentException("Parameter instanceName should be provided.");
+        }
+    }
+
     default void checkRegistrarOptionsNotNull(RegistrarOptions options) {
         if (options == null) {
             throw new IllegalArgumentException("Parameter registrar options should be provided.");
         }
     }
 
-    Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataKeyType> metadata, String ipAddress,
-            int defaultPort);
+    default Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataKeyType> metadata, String ipAddress,
+            int defaultPort) {
+        return registerServiceInstance(serviceName, null, metadata, ipAddress, defaultPort);
+    }
+
+    Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<MetadataKeyType> metadata,
+            String ipAddress, int defaultPort);
 
     default Uni<Void> registerServiceInstance(RegistrarOptions options) {
         checkRegistrarOptionsNotNull(options);
@@ -34,7 +45,28 @@ public interface ServiceRegistrar<MetadataKeyType extends Enum<MetadataKeyType> 
         return registerServiceInstance(options.serviceName(), options.ipAddress(), options.defaultPort());
     }
 
+    @Deprecated
     Uni<Void> deregisterServiceInstance(String serviceName);
+
+    /**
+     * Deregisters a specific service instance identified by instance name.
+     * Implementors SHOULD override this method.
+     * The default implementation falls back to {@link #deregisterServiceInstance(String)},
+     * which may deregister ALL instances of the service.
+     */
+    default Uni<Void> deregisterServiceInstance(String serviceName, String instanceName) {
+        return deregisterServiceInstance(serviceName);
+    }
+
+    /**
+     * Deregisters a specific service instance identified by its IP address and port.
+     * Implementors SHOULD override this method.
+     * The default implementation falls back to {@link #deregisterServiceInstance(String)},
+     * which may deregister ALL instances of the service.
+     */
+    default Uni<Void> deregisterServiceInstance(String serviceName, String ipAddress, int port) {
+        return deregisterServiceInstance(serviceName);
+    }
 
     record RegistrarOptions(String serviceName, String ipAddress, int defaultPort, List<String> tags,
             Map<String, String> metadata) {
--- a/core/src/test/java/io/smallrye/stork/TestServiceRegistrar.java
+++ b/core/src/test/java/io/smallrye/stork/TestServiceRegistrar.java
@@ -19,13 +19,13 @@ public class TestServiceRegistrar implements ServiceRegistrar<TestMetadataKey> {
     }
 
     @Override
-    public Uni<Void> registerServiceInstance(String serviceName, Metadata<TestMetadataKey> metadata, String ipAddress,
-            int defaultPort) {
-        return null;
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<TestMetadataKey> metadata,
+            String ipAddress, int defaultPort) {
+        throw new UnsupportedOperationException();
     }
 
     @Override
     public Uni<Void> deregisterServiceInstance(String serviceName) {
-        return null;
+        throw new UnsupportedOperationException();
     }
 }
--- a/docs/snippets/examples/AcmeServiceRegistrar.java
+++ b/docs/snippets/examples/AcmeServiceRegistrar.java
@@ -16,7 +16,8 @@ public class AcmeServiceRegistrar implements ServiceRegistrar {
 
 
     @Override
-    public Uni<Void> registerServiceInstance(String serviceName, Metadata metadata, String ipAddress, int defaultPort) {
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata metadata, String ipAddress, int defaultPort) {
+        //instanceName is an optional identifier for this specific instance; use it as the registration ID when provided
         //do whatever is needed for registering service instance
         return Uni.createFrom().voidItem();
     }
--- a/service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
+++ b/service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
@@ -1,9 +1,9 @@
 package io.smallrye.stork.serviceregistration.consul;
 
-import static io.smallrye.stork.impl.ConsulMetadataKey.META_CONSUL_SERVICE_ID;
-
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -18,6 +18,8 @@ import io.vertx.core.net.JksOptions;
 import io.vertx.ext.consul.CheckOptions;
 import io.vertx.ext.consul.ConsulClient;
 import io.vertx.ext.consul.ConsulClientOptions;
+import io.vertx.ext.consul.Service;
+import io.vertx.ext.consul.ServiceList;
 import io.vertx.ext.consul.ServiceOptions;
 
 public class ConsulServiceRegistrar implements ServiceRegistrar<ConsulMetadataKey> {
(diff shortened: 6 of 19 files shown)


service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
```diff
@@ -95,16 +97,17 @@ private Uni<Void> registerInstance(String serviceName, String ipAddress, int def
...
                         em.fail(result.cause());
                     } else {
-                        log.info("Instances of service {} has been registered ", serviceName);
+                        log.info("Instance {} of service {} has been registered ", consulId, serviceName);
                         em.complete(result.result());
                     }
                 }));
     }
 
+    @Deprecated
```
cescoffier: I think it should not be deprecated, but removed altogether and justified.

The Consul API treats this argument as a service ID, not a  service name. Before this PR, the ID was equal to the service name, so it worked. Now that the ID is serviceName-ip-port, calling deregisterServiceInstance("my-ervice") will silently fail to deregister an  instance registered with ID my-service-10.96.96.231-8406.

cescoffier: Maybe we could track registered IDs internally and provide a deregisterAll(serviceName) method, or iterate over Consul's catalog to find and deregister matching services.

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
```diff
@@ -209,6 +209,89 @@ void shouldRegisterServiceInstancesWithHealthCheckInConsul() throws InterruptedE
         });
     }
 
+    @Test
```
cescoffier, line 179: Test named shouldRegisterConsulIdDefaultingToServiceName don't make sense anymore (the test name not the content)

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
```diff
@@ -209,6 +209,89 @@ void shouldRegisterServiceInstancesWithHealthCheckInConsul() throws InterruptedE
...
+                                emitter.complete(result.result());
+                            }
+                        }));
+
+        UniAssertSubscriber<ServiceEntryList> subscriber = serviceEntryList
+                .subscribe().withSubscriber(UniAssertSubscriber.create());
+
+        ServiceEntryList item = subscriber.awaitItem().getItem();
+        assertThat(item).isNotNull();
+        assertThat(item.getList()).hasSize(2);
```
cescoffier, line 216: Can we be a bit more precise?

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
```diff
@@ -209,6 +209,89 @@ void shouldRegisterServiceInstancesWithHealthCheckInConsul() throws InterruptedE
...
```
cescoffier, line 307: Can we be more precise?

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
```diff
@@ -67,7 +68,8 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<ConsulMeta
             int defaultPort) {
         checkAddressNotNull(ipAddress);
 
-        String consulId = metadata.getMetadata().isEmpty() ? serviceName
+        String consulId = metadata.getMetadata().isEmpty()
```
cescoffier: This should be moved to a helper method carefully documented.

cescoffier: Also, service names, IP addresses, and the separator are all hyphen-delimited, which makes IDs non-parseable. In `my-service-10.96.96.231-8406`, where does the service name end? This isn't a problem for Consul (it treats IDs as opaque strings), but could matter if you ever need to parse IDs
 back into service names. A different separator (/ or ::) would make the format unambiguous. Basically any character not allowed in a Consul service name.

cescoffier (changes requested):

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
aureamunoz: Yes, you're right, I hadn't thought about it. I've replaced the broken implementation with one that queries Consul's catalog via catalogServiceNodes(serviceName), collects all registered service IDs, and deregisters them in parallel using Uni.combine().all().unis(...).discardItems(). This is more robust.

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
aureamunoz: :+1: the test name was misleading and the scenario is already fully covered by  shouldRegisterMultipleInstancesWithUniqueConsulIds. I've removed it

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
aureamunoz: Yes. Done

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
aureamunoz: Done

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
aureamunoz: Extracted to a private static buildConsulId(serviceName, ipAddress, port) method, used consistently in both registerServiceInstance overloads and deregisterServiceInstance(String, String, int). Aslo changed the separator to `::`.

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
```diff
@@ -95,31 +99,77 @@ private Uni<Void> registerInstance(String serviceName, String ipAddress, int def
...
                     }
                 }));
     }
 
+    /**
+     * Deregisters all instances of the given service from Consul by querying the catalog for their IDs
+     * and deregistering them in parallel. This avoids silent failures that would occur if the service ID
+     * were assumed to match the service name.
+     */
     @Override
```
cescoffier, line 133: The other implementation deprecated this method. I think we need to be consistent.

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -59,12 +60,14 @@ public Uni<Void> registerServiceInstance(RegistrarOptions options) {
...
         return registerApplicationInstance(client, options.serviceName(),
-                options.serviceName(), options.ipAddress(), null,
+                eurekaId, options.ipAddress(), null,
                 options.defaultPort(), null, -1, "UP", options.metadata());
 
     }
 
+    @Deprecated
     @Override
     public Uni<Void> deregisterServiceInstance(String serviceName) {
```
cescoffier, line 79: Consul is doing an unregister all instances with the name, here only one. Is it what we want?

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
```diff
@@ -95,31 +99,77 @@ private Uni<Void> registerInstance(String serviceName, String ipAddress, int def
...
                     } else {
-                        log.info("Instances of service {} has been deregistered ", serviceName);
+                        log.info("Instance {} of service {} has been deregistered", consulId, serviceName);
                         em.complete(result.result());
                     }
                 }));
     }
 
+    // Consul service names cannot contain ':', making '::' an unambiguous separator between name, IP, and port.
+    private static String buildConsulId(String serviceName, String ipAddress, int port) {
```
cescoffier, line 175: Consul doesn't strictly forbid colons in service names, it's DNS that does, and Consul allows non-DNS-compliant names via the API (which we use). This separator choice is pragmatic (no one uses `::` in service names, hopefully), but the comment overstates the guarantee. 

I would add proper javadoc with:
```
// '::' is used as separator; it does not appear in well-formed DNS service names, IP addresses, or port numbers.
```

Same for Eureka.

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
```diff
@@ -57,7 +57,7 @@ void setUp() {
     }
 
     @Test
-    void shouldRegisterConsulIdDefaultingToServiceName() {
+    void shouldRegisterServiceInstancesInConsul() throws InterruptedException {
```
cescoffier, line 60: Do we have a test dereigstering a specific instance while keeping the others?

cescoffier (changes requested):

service-registration/consul/src/test/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrationTest.java
aureamunoz: Added one

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -34,8 +34,13 @@ default Uni<Void> registerServiceInstance(RegistrarOptions options) {
         return registerServiceInstance(options.serviceName(), options.ipAddress(), options.defaultPort());
     }
 
+    @Deprecated
     Uni<Void> deregisterServiceInstance(String serviceName);
 
+    default Uni<Void> deregisterServiceInstance(String serviceName, String ipAddress, int port) {
+        return deregisterServiceInstance(serviceName);
```
cescoffier, line 68: is it really a safe fallback?

I would at least add:

/**
    * Deregisters a specific service instance. Implementors SHOULD override this method.
    * The default implementation falls back to {@link #deregisterServiceInstance(String)},
    * which may deregister ALL instances of the service.
    */

cescoffier (dismissed):
please look at the javadoc for the fallback.

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
aureamunoz: Done

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -59,24 +63,40 @@ public Uni<Void> registerServiceInstance(RegistrarOptions options) {
...
+     * and deregistering them in parallel. This avoids silent failures that would occur if the service ID
+     * were assumed to match the service name.
+     */
     @Override
     public Uni<Void> deregisterServiceInstance(String serviceName) {
         return client.get(path + serviceName)
                 .putHeader("Accept", "application/json;charset=UTF-8")
-                .send().invoke(() -> log.info("Instance found for '{}'", serviceName))
+                .send()
                 .flatMap(item -> {
```
cescoffier, line 83: What happens if the service is not found? Don't we get `null` as body?

cescoffier: I think we would need something like this:
```
if (item.statusCode() == 404) {
          log.info("No application found for '{}'", serviceName);
          return Uni.createFrom().voidItem();
      }
      JsonObject application = item.bodyAsJsonObject().getJsonObject("application");
```

cescoffier (changes requested):

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
aureamunoz: yes, you're right. Done

core/src/test/java/io/smallrye/stork/TestServiceRegistrar.java
```diff
@@ -24,6 +24,12 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<TestMetada
         return null;
     }
 
+    @Override
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<TestMetadataKey> metadata,
+            String ipAddress, int defaultPort) {
+        return null;
```
cescoffier: If it should not be called, throw an UnsupportedOperationException or an AssertionError.

core/src/test/java/io/smallrye/stork/TestServiceRegistrar.java
```diff
@@ -24,6 +24,12 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<TestMetada
...
+    }
+
     @Override
     public Uni<Void> deregisterServiceInstance(String serviceName) {
```
cescoffier, line 28: Same here, if it should not be called, throw an UnsupportedOperationException or an AssertionError.

core/src/test/java/io/smallrye/stork/TestServiceRegistrar.java
```diff
@@ -24,6 +24,12 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<TestMetada
         return null;
```
cescoffier: If it should not be called, throw an UnsupportedOperationException or an AssertionError.

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -34,8 +45,29 @@ default Uni<Void> registerServiceInstance(RegistrarOptions options) {
         return registerServiceInstance(options.serviceName(), options.ipAddress(), options.defaultPort());
     }
 
+    @Deprecated
     Uni<Void> deregisterServiceInstance(String serviceName);
 
+    /**
+     * Deregisters a specific service instance identified by its IP address and port.
```
cescoffier: "identified by its IP address and port" but the method takes `(serviceName, instanceName)` - I believed you copied it.

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -59,24 +63,61 @@ public Uni<Void> registerServiceInstance(RegistrarOptions options) {
...
-                    JsonObject body = item.bodyAsJsonObject();
-                    JsonObject application = body.getJsonObject("application");
-                    JsonObject instance = application.getJsonArray("instance").getJsonObject(0);
-                    return deregisterApplicationInstance(application.getString("name"), instance.getString("instanceId"));
+                    if (item.statusCode() == 404) {
```
cescoffier, line 84: we should guard against 500 and 503 too:
https://stackoverflow.com/questions/75929738/eureka-server-getting-503-service-unavailable-when-accessing-a-endpoint-url

cescoffier (changes requested):

test-utils/src/main/java/io/smallrye/stork/test/TestServiceRegistrarProvider.java
```diff
@@ -71,6 +71,12 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata<TestMetada
             return Uni.createFrom().voidItem();
         }
 
+        @Override
+        public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<TestMetadata> metadata,
+                String ipAddress, int defaultPort) {
+            return null;
```
cescoffier: Should not return `null`.

cescoffier: throw exception or return a Uni.createFrom().voidItem();

cescoffier: Just to explain: returning `null` from a `Uni<Void>` method will cause `NullPointerException` when subscribed.

docs/snippets/examples/AcmeServiceRegistrar.java
```diff
@@ -21,6 +21,11 @@ public Uni<Void> registerServiceInstance(String serviceName, Metadata metadata,
         return Uni.createFrom().voidItem();
     }
 
+    @Override
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata metadata, String ipAddress, int defaultPort) {
+        return null;
```
cescoffier: Same here, do not return null. Returning `null` from a `Uni<Void>` method will cause `NullPointerException` when subscribed.

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -59,24 +63,66 @@ public Uni<Void> registerServiceInstance(RegistrarOptions options) {
...
+                        log.info("No application found for '{}'", serviceName);
+                        return Uni.createFrom().voidItem();
+                    }
+                    if (item.statusCode() == 500 || item.statusCode() == 503) {
```
cescoffier: I would say that if we reach this point, we could say: item.statusCode() >= 400

cescoffier (changes requested):

test-utils/src/main/java/io/smallrye/stork/test/TestServiceRegistrarProvider.java
aureamunoz: yes, I've seen it :-) thanks

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
aureamunoz: ok, good idea

aureamunoz: Thinking a bit more about the Eureka error handling, I think it makes more sense to keep the 404 case handled separately to have idempotent behavior, like DELETE in APIs such as k8s, if you try to delete something that's not there, it doesn't fail. The rest I've grouped together though.

api/src/main/java/io/smallrye/stork/api/Metadata.java
```diff
@@ -97,6 +98,18 @@ public EnumMap<T, Object> getMetadata() {
...
     }
 
+    /**
+     * @return the metadata as a {@code Map<String, String>}, using each key's name and converting values via
+     *         {@code toString()}
+     */
+    public Map<String, String> asMap() {
+        Map<String, String> result = new HashMap<>();
+        for (Map.Entry<T, Object> entry : metadata.entrySet()) {
+            result.put(entry.getKey().getName(), entry.getValue().toString());
```
cescoffier: if the value is `null`, this will throw an NPE.

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -235,14 +235,27 @@ public Uni<Void> registerInstance(String serviceName, String ipAddress, int port
         return serviceRegistrar.registerServiceInstance(serviceName, ipAddress, port);
     }
 
+    public Uni<Void> registerInstance(String serviceName, String instanceName, String ipAddress, int port) {
```
cescoffier, line 255: Missing javadoc on the new methods.

cescoffier (changes requested):
At least the NPE need to be fixed.

cescoffier (approved):

PR #1244, opened 2026-05-18
fix: bump to Mutiny Vert.x bindings 3.22.2

jponge:
This fixes a Java module descriptor as the Mutiny bindings now have an optional runtime dependency on the Vert.x code generator code annotations.

Supersedes #1228


aureamunoz (approved):
Thank you!

Issue #1253, opened 2026-06-02
Remove redundant serviceName parameter from Service registration/deregistration methods


aureamunoz:
## Problem

When registering a service instance programmatically, the service name must be passed twice — once to obtain the `Service` object and again to the registration method:

```java
stork.getService("band-service").registerInstance("band-service", "Slash", "localhost", 9000);
```

This is redundant because `Service` already stores its `serviceName` (received at construction time) and could use it internally.

The same redundancy applies to all registration/deregistration methods on `Service`:
- `registerInstance(String serviceName, String ipAddress, int port)`
- `registerInstance(String serviceName, String instanceName, String ipAddress, int port)`
- `deregisterServiceInstance(String serviceName)`
- `deregisterServiceInstance(String serviceName, String ipAddress, int port)`
- `deregisterServiceInstance(String serviceName, String instanceName)`

## Proposed solution

Remove the redundant `serviceName` parameter from all registration/deregistration methods on `Service`, using `this.serviceName` internally. Also expose the new `registerServiceInstance` variant with tags support.

### New methods on `Service`

```java
registerInstance(String ipAddress, int port)
registerInstance(String instanceName, String ipAddress, int port)
registerInstance(String instanceName, List<String> tags, String ipAddress, int port)
registerInstance(ServiceRegistrar.RegistrarOptions options)
deregisterServiceInstance()
deregisterServiceInstance(String instanceName)
deregisterServiceInstance(String ipAddress, int port)
```

### Removed methods (previously deprecated)

All former overloads that took `serviceName` as a parameter have been removed.

## Scope

- **Changed:** `Service.java` in the `api` module — old methods removed, new methods added.
- **Changed:** `revapi.json` — added ignore entry for the new `registerServiceInstance` with tags on `ServiceRegistrar`.
- **Changed:** `docs/snippets/examples/AcmeServiceRegistrar.java` — fixed to use parameterized `ServiceRegistrar<Metadata.DefaultMetadataKey>` and implement all abstract methods.
- **Unchanged:** `ServiceRegistrar` interface (SPI) — implementations (Consul, Eureka, static-list) are not affected.
- **Unchanged:** Quarkus, MicroProfile, and Spring Boot integrations.

## Target API after the change

```java
stork.getService("band-service").registerInstance("Slash", "localhost", 9000);
stork.getService("band-service").registerInstance("Slash", List.of("v1.0", "canary"), "localhost", 9000);
[...]

PR #1254, opened 2026-06-02
Improve api registration

aureamunoz:
Related to #1253

### DIFF
--- a/api/src/main/java/io/smallrye/stork/api/Service.java
+++ b/api/src/main/java/io/smallrye/stork/api/Service.java
@@ -234,26 +234,65 @@ public class Service {
     /**
      * Registers a service instance identified by its IP address and port.
      *
-     * @param serviceName the name of the service
      * @param ipAddress the IP address of the instance
      * @param port the port of the instance
      * @return a {@link Uni} completing when the registration is done
      */
-    public Uni<Void> registerInstance(String serviceName, String ipAddress, int port) {
+    public Uni<Void> registerInstance(String ipAddress, int port) {
+        if (serviceRegistrar == null) {
+            return Uni.createFrom()
+                    .failure(new UnsupportedOperationException("This service has no service registrar configured."));
+        }
         return serviceRegistrar.registerServiceInstance(serviceName, ipAddress, port);
     }
 
     /**
      * Registers a service instance with an explicit instance name used as the registration ID.
      *
-     * @param serviceName the name of the service
      * @param instanceName the unique identifier for this instance in the registry
      * @param ipAddress the IP address of the instance
      * @param port the port of the instance
      * @return a {@link Uni} completing when the registration is done
      */
-    public Uni<Void> registerInstance(String serviceName, String instanceName, String ipAddress, int port) {
+    public Uni<Void> registerNamedInstance(String instanceName, String ipAddress, int port) {
+        if (serviceRegistrar == null) {
+            return Uni.createFrom()
+                    .failure(new UnsupportedOperationException("This service has no service registrar configured."));
+        }
         return serviceRegistrar.registerServiceInstance(serviceName, instanceName, Metadata.empty(), ipAddress, port);
+    }
+
+    /**
+     * Registers a service instance with tags.
+     *
+     * @param tags the tags to associate with this instance
+     * @param ipAddress the IP address of the instance
+     * @param port the port of the instance
+     * @return a {@link Uni} completing when the registration is done
+     */
+    public Uni<Void> registerInstance(List<String> tags, String ipAddress, int port) {
+        if (serviceRegistrar == null) {
+            return Uni.createFrom()
+                    .failure(new UnsupportedOperationException("This service has no service registrar configured."));
+        }
+        return serviceRegistrar.registerServiceInstance(serviceName, tags, Metadata.empty(), ipAddress, port);
+    }
+
+    /**
+     * Registers a service instance with an explicit instance name and tags.
+     *
+     * @param instanceName the unique identifier for this instance in the registry
+     * @param tags the tags to associate with this instance
+     * @param ipAddress the IP address of the instance
+     * @param port the port of the instance
+     * @return a {@link Uni} completing when the registration is done
+     */
+    public Uni<Void> registerNamedInstance(String instanceName, List<String> tags, String ipAddress, int port) {
+        if (serviceRegistrar == null) {
+            return Uni.createFrom()
+                    .failure(new UnsupportedOperationException("This service has no service registrar configured."));
+        }
+        return serviceRegistrar.registerServiceInstance(serviceName, instanceName, tags, Metadata.empty(), ipAddress, port);
     }
 
     /**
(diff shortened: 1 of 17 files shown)


service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -105,12 +105,20 @@ public Uni<Void> deregisterServiceInstance(String serviceName) {
     @Override
     public Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<EurekaMetadataKey> metadata,
             String ipAddress, int defaultPort) {
+        return registerServiceInstance(serviceName, instanceName, List.of(), metadata, ipAddress, defaultPort);
+    }
+
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
```
cescoffier, line 113: Missing @Override.

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -234,28 +234,30 @@ public ServiceRegistrar getServiceRegistrar() {
     /**
```
cescoffier, line 234: You are removing method that were not deprecated. Consider deprecating them first.

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -38,6 +38,9 @@ default Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataK
     Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<MetadataKeyType> metadata,
             String ipAddress, int defaultPort);
 
+    Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
```
cescoffier: This is also a breaking change. Consider adding a default implementation:

```
default Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
        Metadata<MetadataKeyType> metadata, String ipAddress, int defaultPort) {
    return registerServiceInstance(serviceName, instanceName, metadata, ipAddress, defaultPort);
}
```

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -234,28 +234,30 @@ public ServiceRegistrar getServiceRegistrar() {
...
      * @param ipAddress the IP address of the instance
      * @param port the port of the instance
      * @return a {@link Uni} completing when the registration is done
      */
-    public Uni<Void> registerInstance(String serviceName, String instanceName, String ipAddress, int port) {
+    public Uni<Void> registerInstance(String instanceName, String ipAddress, int port) {
         return serviceRegistrar.registerServiceInstance(serviceName, instanceName, Metadata.empty(), ipAddress, port);
     }
 
+    public Uni<Void> registerInstance(String instanceName, List<String> tags, String ipAddress, int port) {
```
cescoffier: Missing javadoc

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -38,6 +38,9 @@ default Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataK
...
+            Metadata<MetadataKeyType> metadata, String ipAddress, int defaultPort);
+
     default Uni<Void> registerServiceInstance(RegistrarOptions options) {
```
cescoffier, line 51: Should call the variant using tags.

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -38,6 +38,9 @@ default Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataK
     Uni<Void> registerServiceInstance(String serviceName, String instanceName, Metadata<MetadataKeyType> metadata,
```
cescoffier, line 43: What about `registerInstance(List<String> tags, String ipAddress, int port)` ?

cescoffier (changes requested):

api/src/main/java/io/smallrye/stork/api/Service.java
aureamunoz: You're right. The thing is, registerInstance(String, String, int) wasn't truly removed, its signature is unchanged. What changed is the semantics of the first parameter: from serviceName to instanceName. We can't deprecate the old and add the new simultaneously because they have the same erasure. This means existing callers passing serviceName would silently have it interpreted as instanceName. To avoid that, I could remove the 3-arg overload entirely and keep only registerInstance(String ipAddress, int port)  + registerInstance(RegistrarOptions). WDYT?

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
aureamunoz: Added

api/src/main/java/io/smallrye/stork/api/Service.java
cescoffier: The problem is that it can really break application badly. We may have to introduce a new method to avoid this.

cescoffier: I think we have:

- `registerInstance(String serviceName, String ipAddress, int port)` -> replaced by `registerInstance(String ipAddress, int port)`
   - `registerInstance(String serviceName, String instanceName, String ipAddress, int port)` -> replaced by `registerInstance(String instanceName, String ipAddress, int port)`
   - `deregisterServiceInstance(String serviceName)` -> replaced by `deregisterServiceInstance()`
   - `deregisterServiceInstance(String serviceName, String ipAddress, int port)` -> replaced by `deregisterServiceInstance(String ipAddress, int port)`
   - `deregisterServiceInstance(String serviceName, String instanceName)` -> replaced by `deregisterServiceInstance(String instanceName)`

   This is a binary-incompatible change for any downstream code calling these methods. Any Quarkus extension or user code using `service.registerInstance("my-service", "10.0.0.1", 8080)` will get a `NoSuchMethodError` at runtime.

Do I miss something?

service-registration/static-list/src/main/java/io/smallrye/stork/serviceregistration/staticlist/StaticListServiceRegistrar.java
```diff
@@ -23,7 +25,16 @@ public StaticListServiceRegistrar(StaticRegistrarConfiguration config, String se
...
     public Uni<Void> registerServiceInstance(String serviceName, String instanceName,
             Metadata<Metadata.DefaultMetadataKey> metadata, String ipAddress, int defaultPort) {
+        return registerServiceInstance(serviceName, instanceName, List.of(), metadata, ipAddress, defaultPort);
+    }
+
+    @Override
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
+            Metadata<Metadata.DefaultMetadataKey> metadata,
+            String ipAddress,
+            int defaultPort) {
```
cescoffier, line 35: Tags should be passed there.

service-registration/eureka/src/main/java/io/smallrye/stork/serviceregistration/eureka/EurekaServiceRegistrar.java
```diff
@@ -105,12 +105,21 @@ public Uni<Void> deregisterServiceInstance(String serviceName) {
...
+    @Override
+    public Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
+            Metadata<EurekaMetadataKey> metadata,
+            String ipAddress,
+            int defaultPort) {
         checkAddressNotNull(ipAddress);
         String eurekaId = instanceName != null && !instanceName.isEmpty()
                 ? instanceName
                 : buildEurekaId(serviceName, ipAddress, defaultPort);
         return registerApplicationInstance(client, serviceName, eurekaId, ipAddress, null, defaultPort, null, -1, "UP",
```
cescoffier, line 125: Tags must be passed here.

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
```diff
@@ -38,11 +38,17 @@ default Uni<Void> registerServiceInstance(String serviceName, Metadata<MetadataK
...
+            Metadata<MetadataKeyType> metadata, String ipAddress, int defaultPort) {
+        return registerServiceInstance(serviceName, instanceName, metadata, ipAddress, defaultPort);
+    }
+
     default Uni<Void> registerServiceInstance(RegistrarOptions options) {
         checkRegistrarOptionsNotNull(options);
         checkAddressNotNull(options.ipAddress());
 
-        return registerServiceInstance(options.serviceName(), options.ipAddress(), options.defaultPort());
+        return registerServiceInstance(options.serviceName(), null, options.tags(), Metadata.empty(), options.ipAddress(),
```
cescoffier: We need to be careful passing `null` as tags, I would prefer List.of()

cescoffier: `options.tags() != null ? options.tags() : List.of()`

cescoffier: Why Metadata.empty() and not options.metadata() ?

api/src/main/java/io/smallrye/stork/api/Service.java
```diff
@@ -234,28 +234,51 @@ public ServiceRegistrar getServiceRegistrar() {
...
      * Registers a service instance identified by its IP address and port.
      *
-     * @param serviceName the name of the service
      * @param ipAddress the IP address of the instance
      * @param port the port of the instance
      * @return a {@link Uni} completing when the registration is done
      */
-    public Uni<Void> registerInstance(String serviceName, String ipAddress, int port) {
+    public Uni<Void> registerInstance(String ipAddress, int port) {
         return serviceRegistrar.registerServiceInstance(serviceName, ipAddress, port);
```
cescoffier, line 246: servie registar can be null, we need to be more careful here. Otherwise you get a cryptic NPE.

service-registration/consul/src/main/java/io/smallrye/stork/serviceregistration/consul/ConsulServiceRegistrar.java
```diff
@@ -69,12 +69,20 @@ public Uni<Void> registerServiceInstance(RegistrarOptions options) {
...
+            Metadata<ConsulMetadataKey> metadata,
+            String ipAddress,
+            int defaultPort) {
         checkAddressNotNull(ipAddress);
 
         // Use the explicit ID when provided; otherwise generate a unique one.
         String consulId = instanceName != null && !instanceName.isEmpty() ? instanceName
                 : buildConsulId(serviceName, ipAddress, defaultPort);
-        return registerInstance(serviceName, ipAddress, defaultPort, consulId, List.of(), Collections.emptyMap());
+        return registerInstance(serviceName, ipAddress, defaultPort, consulId, tags, Collections.emptyMap());
```
cescoffier, line 85: We need tests covering tags.

test-utils/src/main/java/io/smallrye/stork/test/TestServiceRegistrarProvider.java
```diff
@@ -69,9 +69,18 @@ public TestServiceRegistrar(TestSrRegistrarConfiguration config, String serviceR
...
                 String ipAddress, int defaultPort) {
+            return registerServiceInstance(serviceName, instanceName, List.of(), metadata, ipAddress, defaultPort);
+        }
+
+        @Override
+        public Uni<Void> registerServiceInstance(String serviceName, String instanceName, List<String> tags,
+                Metadata<TestMetadata> metadata,
+                String ipAddress,
+                int defaultPort) {
             registrations.add(new Registration(serviceRegistrarName, config, metadata, serviceName, instanceName, ipAddress,
```
cescoffier: Where are the tags?

cescoffier (changes requested):

api/src/main/java/io/smallrye/stork/api/ServiceRegistrar.java
aureamunoz: RegistrarOptions.metadata() is a Map<String, String>, incompatible with Metadata<MetadataKeyType>. Implementations that need it (Consul, Eureka) already override this method and read options.metadata() directly.

aureamunoz:
Closing before branch swap. Will reopen against new main.

PR #1275, opened 2026-06-19
Fix blocking event loop k8s

### LINKED ISSUE #1274: Kubernetes service discovery blocks Vert.x event loop during informer setup
KubernetesServiceDiscovery.configureInformerOnce() calls op.inform() (a blocking Kubernetes API call) directly on the Vert.x event loop thread. This blocks the event loop and causes request timeouts when Stork is used with reactive clients (e.g. gRPC via StorkGrpcChannel).
The problem happens because vertx.executeBlocking() in shouldUseEndpointSlices() dispatches its result back to the event loop. The flatMap callback then runs on the event loop, and configureInformerOnce() → op.inform() blocks it.
A second blocking point exists when use-cluster-ip=true: the constructor calls configureServicesInformer() synchronously, which also calls op.inform().
Expected behavior: Informer setup should run on a worker thread, not on the event loop.
Actual behavior: The event loop is blocked during informer initialization, causing BlockedThreadChecker warnings and request timeouts.

Related to #1231

aureamunoz:
fixes #1274

### DIFF
--- a/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
+++ b/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
@@ -126,11 +126,8 @@ public class KubernetesServiceDiscovery extends CachingServiceDiscovery {
             LOGGER.warn("Both 'use-cluster-ip' and 'use-endpoint-slices' are enabled for service '{}'. "
                     + "'use-cluster-ip' takes precedence; 'use-endpoint-slices' will be ignored.", serviceName);
         }
-        if (useClusterIp) {
-            configureServicesInformer();
-        }
-        // the endpoints/slices informer is configured lazily on the first fetchNewServiceInstances call,
-        // once the Uni resolves which API the cluster supports
+        // all informers are configured lazily on the first fetchNewServiceInstances call,
+        // to avoid blocking I/O during construction
     }
 
     /**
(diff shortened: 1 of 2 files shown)


aureamunoz:
fyi @jdussouillez

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -247,32 +244,48 @@ public void invalidate() {
     @Override
     public Uni<List<ServiceInstance>> fetchNewServiceInstances(List<ServiceInstance> previousInstances) {
         if (useClusterIp) {
-            return fetchServiceClusterIp().onItem()
-                    .transform(clusterIps -> fromClusterIpServicesToStorkServiceInstances(clusterIps, previousInstances))
-                    .invoke(() -> invalidated.set(false));
+            Uni<Boolean> informerSetup = configureInformerOnce(this::configureServicesInformer);
```
cescoffier: `configureInformerOnce(this::configureServicesInformer)` and `fetchServiceClusterIp()` run in parallel. The `fetchServiceClusterIp()` reads directly from the Kubernetes API (not from the informer cache), so this parallelism is safe for fetching. 
However, the informer is what drives cache invalidation. If the informer hasn't finished configuring when the fetch completes, there's a window during which a change event could be missed: the fetch returns stale data that gets cached, but the informer isn't watching yet to trigger invalidation.

I think we need to consider sequencing the informer setup before the fetch (on the worker thread), rather than running them in parallel.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -247,32 +244,48 @@ public void invalidate() {
...
-            if (useSlices) {
-                configureSlicesInformer();
-            } else {
-                configureEndpointsInformer();
-            }
+            return executeOnWorkerThread(() -> {
+                configurator.run();
+                return null;
+            }).replaceWith(true)
+                    .onFailure().invoke(th -> LOGGER.warn("Failed to configure informer", th))
```
cescoffier: Merge the 2 invoke.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
```diff
@@ -247,32 +244,48 @@ public void invalidate() {
...
+    private Uni<Boolean> configureInformerOnce(Runnable configurator) {
         if (informerConfigured.compareAndSet(false, true)) {
-            if (useSlices) {
-                configureSlicesInformer();
-            } else {
-                configureEndpointsInformer();
-            }
+            return executeOnWorkerThread(() -> {
+                configurator.run();
+                return null;
```
cescoffier: Why not use something like `executeOnWorkerThread(Runnable runnable)`?

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryEventLoopTest.java
```diff
@@ -0,0 +1,258 @@
...
+import io.fabric8.kubernetes.client.server.mock.EnableKubernetesMockClient;
+import io.fabric8.kubernetes.client.server.mock.KubernetesMockServer;
+import io.smallrye.stork.api.ServiceInstance;
+import io.vertx.core.Vertx;
+
+/**
+ * Verifies that the Kubernetes service discovery informer setup does not block
+ * the Vert.x event loop.
+ *
+ * @see <a href="https://github.com/smallrye/smallrye-stork/issues/1231">smallrye-stork#1231</a>
```
cescoffier: I think #1274 would be better.

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryEventLoopTest.java
```diff
@@ -0,0 +1,258 @@
...
+        KubernetesConfiguration config = new KubernetesConfiguration(Map.of(
+                "k8s-host", k8sMasterUrl,
+                "k8s-namespace", defaultNamespace,
+                "use-cluster-ip", "true"));
+
+        Vertx vertx = Vertx.vertx();
+        try {
+            KubernetesServiceDiscovery discovery = new KubernetesServiceDiscovery("svc", config, vertx);
+
+            List<ServiceInstance> first = discovery.getServiceInstances().await().indefinitely();
```
cescoffier: Use duration for both lines.

cescoffier (changes requested):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: Ok, I see

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryEventLoopTest.java
aureamunoz: done

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: I've added the executeOnWorkerThread(Runnable) overload, but I'm not entirely sure I understood your suggestion correctly. configureInformerOnce already receives a Runnable,  the new overload just moves the return null wrapping from the call site into a method.

cescoffier (approved):

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
cescoffier: Thé idea is the possibility to use a method reference.

cescoffier:
It was not squashed before merging.

aureamunoz:
At this point it's already part of the last release, so there's nothing we can really do for it

PR #1277, opened 2026-06-19
Fix ignoring allNamespaces when fetching slices

### LINKED ISSUE #1276: Kubernetes EndpointSlice discovery ignores k8s-namespace=all queries only the default namespace
When using EndpointSlice-based service discovery with k8s-namespace=all, Stork queries only the configured namespace instead of querying across all namespaces.

aureamunoz:
Fixes #1276

### DIFF
--- a/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
+++ b/service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
@@ -295,7 +295,7 @@ public class KubernetesServiceDiscovery extends CachingServiceDiscovery {
         return executeOnWorkerThread(() -> {
             if (allNamespaces) {
                 return client.discovery().v1().endpointSlices()
-                        .inNamespace(namespace)
+                        .inAnyNamespace()
                         .withLabel(SERVICE_SELECTOR, application)
                         .list().getItems();
             } else {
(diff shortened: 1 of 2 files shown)


service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
cescoffier, line 381: Don't you think that instead of "namespace", you should do something like slice.getNamespace (or whatever).

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryWithSlicesTest.java
```diff
@@ -419,6 +419,42 @@ void shouldReturnNoInstancesWhenPortNameMatchesNothing() {
...
+        Service service = stork.getService(serviceName);
+        service.getServiceDiscovery().getServiceInstances()
+                .onFailure().invoke(th -> fail("Failed to get service instances from Kubernetes (EndpointSlices)", th))
+                .subscribe().with(instances::set);
+
+        await().atMost(Duration.ofSeconds(5))
+                .until(() -> instances.get() != null);
+
+        assertThat(instances.get()).hasSize(4);
+        assertThat(instances.get().stream().map(ServiceInstance::getPort)).allMatch(p -> p == 8080);
```
cescoffier, line 455: Can we check the namespace metadata?

cescoffier (dismissed):
Approved, but please look at my comments.

service-discovery/kubernetes/src/main/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscovery.java
aureamunoz: yes. good idea, done

service-discovery/kubernetes/src/test/java/io/smallrye/stork/servicediscovery/kubernetes/KubernetesServiceDiscoveryWithSlicesTest.java
aureamunoz: yes, done

cescoffier (approved):

